PR #2, opened 2020-02-05
 chore: add split repo templates

plamut:
This is step 6 of the repo split procedure.


plamut:
I'm not sure why the commit message check complains, maybe there are some non-printable characters hiding in the commit message or something ...

PR #30, opened 2020-03-04
Use ContainerEngineHandler in Cloud Run environments and include `trace` in logged messages.

jakebiesinger-onduo:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
      - https://issuetracker.google.com/issues/150734081
      - see also #6 and #7 
- [x] Ensure the tests and linter pass

This PR attempts to improve the logging situation for Cloud Run environments by

1. Attempting to read the `trace` associated with common python web app environments (reusing existing logic in use by appengine)
    - This follows the instructions outlined [here](https://cloud.google.com/run/docs/logging#container-logs)
2. Attempting to guess when we're in a Cloud Run environment (similar to appengine + GKE) and selecting the `ContainerEngineHandler` if we are. The json payload written by that handler  includes the logging `level`, whereas the standard `CloudLoggingHandler` drops that information.
     - One concern: the environment variables feel a little brittle, even though that's how AppEngine is solving this problem. If there's, say, some compute metadata we could grab instead to verify we're in Cloud Run land, that seems much cleaner.

I'm happy to update tests and docs if this seems like a good approach to the problem.


jakebiesinger-onduo:
BTW, the only way I could test this change is to define my own Handler, basically copy-pasting the code for `ContainerEngineHandler`. Something like:

```python
from google.cloud.logging.handlers._helpers import format_stackdriver_json, get_trace_id
from google.cloud.logging._helpers import retrieve_metadata_server
import json

logging_client = google.cloud.logging.Client()


_TRACE_ID_LABEL = "logging.googleapis.com/trace"
_PROJECT_ID = "project/project-id"
"""Attribute in metadata server for the current project-id."""


class ContainerEngineHandler(logging.StreamHandler):
  """Handler to format log messages the format expected by GKE fluent.

  This handler is written to format messages for the Google Container Engine
  (GKE) fluentd plugin, so that metadata such as log level are properly set.

  :type name: str
  :param name: (optional) the name of the custom log in Stackdriver Logging.

  :type stream: file-like object
  :param stream: (optional) stream to be used by the handler.
  """

  def __init__(self, name=None, stream=None):
    super(ContainerEngineHandler, self).__init__(stream=stream)
    self.name = name

    self.project_id = retrieve_metadata_server(_PROJECT_ID)

  def get_gke_labels(self):
    """Return the labels for GKE app.

    If the trace ID can be detected, it will be included as a label.
    Currently, no other labels are included.

    :rtype: dict
    :returns: Labels for GKE app.
    """
    gke_labels = {}

    trace_id = get_trace_id()
    if trace_id is not None:
      gke_labels[_TRACE_ID_LABEL] = "projects/{}/traces/{}".format(
        self.project_id, trace_id
      )

    return gke_labels

  def format(self, record):
    """Format the message into JSON expected by fluentd.

    :type record: :class:`~logging.LogRecord`
    :param record: the log record

    :rtype: str
    :returns: A JSON string formatted for GKE fluentd.
    """
    message = super(ContainerEngineHandler, self).format(record)
    gke_labels = self.get_gke_labels()
    values = json.loads(format_stackdriver_json(record, message))
    values.update(**gke_labels)
    return json.dumps(values)



handler = ContainerEngineHandler()
logging.basicConfig(level=logging.INFO, handlers=[handler])

...

@app.route('/', methods=['GET'])
def handle_thing():
  logging.error('test thing')  # <-- sets the `level` and `trace` successfully

```

jakebiesinger-onduo:
Howdy! Friendly ping?

jakebiesinger-onduo:
Moar friendly pings!

google/cloud/logging/handlers/container_engine.py
```diff
@@ -21,7 +21,12 @@
 
 import logging.handlers
 
-from google.cloud.logging.handlers._helpers import format_stackdriver_json
+from google.cloud.logging.handlers._helpers import format_stackdriver_json, get_trace_id
```
tseaver, line 24: One import per line, please.
```suggestion
from google.cloud.logging.handlers._helpers import format_stackdriver_json
from google.cloud.logging.handlers._helpers import get_trace_id
```

google/cloud/logging/handlers/container_engine.py
```diff
@@ -41,6 +46,27 @@ def __init__(self, name=None, stream=None):
         super(ContainerEngineHandler, self).__init__(stream=stream)
         self.name = name
```
tseaver, line 48: Tighten up:
```suggestion
```

Daniel-Sanche:
Hey @jakebiesinger-onduo, thanks for putting this together. I'm going to start adding some improvements to this library over the next few weeks, and improving Cloud Run integration is high up on my priorities. Are you still interested in working on this PR? Or should I take over?

jakebiesinger-onduo:
Jake has left Onduo so feel free to take over!

Daniel-Sanche:
I'm going to close this PR in that case, but I'm hoping to come back to this feature soon

PR #73, opened 2020-10-15
fix: add default filter settings to list_entries

### LINKED ISSUE #71: list_logs hangs
The README gives a sample for how to list logs off of GCP:

```
from google.cloud import logging
client = logging.Client()
logger = client.logger('log_name')
for entry in logger.list_entries():
    print(entry.payload)
```

Running this code, `logger.list_entries()` seems to hang for ~5 minutes, even just to return a single entry

The profiler says it seems to all be inside ` {method 'next_event' of 'grpc._cython.cygrpc.SegregatedCall' objects}`

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       18  290.233   16.124  290.236   16.124 {method 'next_event' of 'grpc._cython.cygrpc.SegregatedCall' objects}
    33/32    0.063    0.002    0.066    0.002 {built-in method _imp.create_dynamic}
      529    0.043    0.000    0.043    0.000 {built-in method marshal.loads}
1901/1889    0.023    0.000    0.115    0.000 {built-in method builtins.__build_class__}
 1166/231    0.018    0.000    0.049    0.000 sre_parse.py:493(_parse)
     3802    0.018    0.000    0.018    0.000 {built-in method posix.stat}

```

daniel-sanche:
Fixes https://github.com/googleapis/python-logging/issues/71 🦕

The `list_entries` functions currently have no filters applied, and the default settings take ~5 minutes for a single query. This PR matches the functionality of `gcloud logging read` by adding a default 24 hour filter to the query. If the user provides their own filter statement, the two statements will be appended, unless they explicitly set their own timestamp value

### DIFF
--- a/google/cloud/logging/_helpers.py
+++ b/google/cloud/logging/_helpers.py
@@ -54,9 +54,6 @@ _NORMALIZED_SEVERITIES = {
     logging.NOTSET: LogSeverity.DEFAULT,
 }
 
-_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f%z"
-"""Time format for timestamps used in API"""
-
 METADATA_URL = "http://metadata.google.internal./computeMetadata/v1/"
 METADATA_HEADERS = {"Metadata-Flavor": "Google"}
 
@@ -144,7 +141,7 @@ def _add_defaults_to_filter(filter_):
 
     # By default, requests should only return logs in the last 24 hours
     yesterday = datetime.now(timezone.utc) - timedelta(days=1)
-    time_filter = 'timestamp>="%s"' % yesterday.strftime(_TIME_FORMAT)
+    time_filter = 'timestamp>="%s"' % yesterday.isoformat()
     if filter_ is None:
         filter_ = time_filter
     elif "timestamp" not in filter_.lower():
--- a/google/cloud/logging/_helpers.py
+++ b/google/cloud/logging/_helpers.py
@@ -16,6 +16,10 @@
 
 import logging
 
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
+
 import requests
 
 from google.cloud.logging.entries import LogEntry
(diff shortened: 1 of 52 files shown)


google/cloud/logging/_helpers.py
```diff
@@ -50,6 +54,9 @@ class LogSeverity(object):
     logging.NOTSET: LogSeverity.DEFAULT,
 }
 
+_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f%z"
```
busunkim96, line 57: Does this need to be kept private? Might be useful for folks who are constructing their own string.

busunkim96 (approved):

google/cloud/logging/_helpers.py
Daniel-Sanche: Maybe, but I'd prefer to be conservative about what we expose publicly. I'd prefer to wait to see if there's demand, or I understand the problem space a bit better

google/cloud/logging/logger.py
```diff
@@ -242,6 +243,7 @@ def list_entries(
         :param filter_:
             a filter expression. See
             https://cloud.google.com/logging/docs/view/advanced_filters
+            By default, a 24 hour filter is applied.
```
tseaver, line 246: Do we need to flag this as a breaking change?

noxfile.py
```diff
@@ -97,7 +97,7 @@ def default(session, django_dep=('django',)):
     )
 
 
-@nox.session(python=['2.7', '3.5', '3.6', '3.7'])
```
tseaver, line 100: Dropping Python 2.7 support is *definitely* a breaking change.

Daniel-Sanche: Sorry, still getting used to the conventional commit style. Yes, this PR introduces a breaking change ([more discussion](https://github.com/googleapis/python-logging/pull/62) on the release PR). How do you think we should proceed since it was already merged without the breaking label?

tseaver: Given that we haven't removed the `Python :: 2` entry from `setup.py` yet, we can repair the breakage by doing that in another PR.

nicoleczhu:
I'm working on handling this in Go lib as well, but realized that checking `"timestamp" not in filter_.lower():`will catch if user is filtering for a payload substring that contains "timestamp" 

Very edge case but may lead to an issue report down the line

Daniel-Sanche:
Hmm good point. I opened https://github.com/googleapis/python-logging/issues/86 to discuss the issue. Thanks!

Issue #76, opened 2020-10-19
Docs are misleading


Daniel-Sanche:
The [documentation on googleapis.dev](https://googleapis.dev/python/logging/latest/index.html) is misleading. It references two versions, with the autogenerated code documented in `v2` and the handwritten layer in `v1`. In actuality, there is only [one version of the API published](https://github.com/googleapis/googleapis/tree/master/google/logging) (`v2`), and the autogenerated layer probably shouldn't be advertised to end-users.

PR #77, opened 2020-10-19
fix: improve docs

daniel-sanche:
**Problem**: 
The generated docs on [googleapis.dev](https://googleapis.dev/python/logging/latest/index.html) are miseading. They advertise two versions of the API, when we actually only publish v2. It also makes it seem like the autogenerated library is the newer version and the handwritten layer is deprecated

**This PR**:
- removed the autogenerated layer from the generated documentation at [googleapis.dev](https://googleapis.dev/python/logging/latest/index.html)
- removed the misleading `v1` version from [googleapis.dev](https://googleapis.dev/python/logging/latest/index.html)
- improved the code samples on the README to use the hand-written layer instead of autogenerated code
- changed branding in the docs from Stackdriver Logging to Cloud Logging and Container Engine to Kubernetes Engine
- got the Kokoro docs-presubmit check to turn green

This should greatly reduce customer confusion

Fixes #https://github.com/googleapis/python-logging/issues/76 🦕


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
-# Changelog
+Changelog
```
busunkim96, line 1: This is a markdown file - was the change to the header intentional?

README.rst
```diff
@@ -77,57 +79,40 @@ Windows
...
-    """
-    e = logging_v2.types.LogEntry(
-        log_name="projects/[PROJECT_ID]/logs/test-logging", # optional
-        resource=resource, # optional
-        text_payload="this is a log statement")
-
-    entries = [e]
-    response = client.write_log_entries(entries)
+    # Imports the Cloud Logging client library
+    import google.cloud.logging
```
busunkim96, line 87: I prefer not to have any codeblocks in the README. There's no good way to test them and it seems like a matter of time until they fall out of date.

Daniel-Sanche: Good point. My intention was to remove the outdated samples, but I'm fine with removing them altogether. Done

CHANGELOG.md
Daniel-Sanche: Good question, I don't think it was intentional? I removed it and it works fine

busunkim96 (approved):

PR #83, opened 2020-10-29
fix: move manual code to logging_v2

busunkim96:
In preparation for moving to the microgenerator, move all manual code into `logging_v2` and have `logging` serve as an alias.

Note: This is targeting the `microgen` branch, not `master`. I'll continue to make PRs to `microgen` to make the reviews more manageable.


google/cloud/logging_v2/__init__.py
```diff
@@ -42,4 +53,6 @@ class MetricsServiceV2Client(metrics_service_v2_client.MetricsServiceV2Client):
     "LoggingServiceV2Client",
     "ConfigServiceV2Client",
     "MetricsServiceV2Client",
+    "ASCENDING",
+    "DESCENDING",
```
Daniel-Sanche, line 57: Isn't `logging` meant to be an alias of `logging_v2`? Right now it looks like they expose different modules. I can't do `logging.logger`, but I can do `logging_v2.logger`

busunkim96: Oops, I messed this up. 

I *think* where I'd like to go with this is what Firestore has:

[`google/cloud/firestore.py`](https://github.com/googleapis/python-firestore/blob/master/google/cloud/firestore.py) (shim) 
[`google/cloud/firestore_v1/__init__.py`](https://github.com/googleapis/python-firestore/blob/master/google/cloud/firestore.py)

The class `Client` (not the module `client`) is placed in `__all__`, so you can import it under `logging_v2`/`logging`. 

Does that sound alright to you? 
```py
from google.cloud import logging

client = logging.Client()
metric = logging.Metric()
```

This is different from what I have at the current commit (exporting modules) which I don't think makes sense for anything but `handlers`.
```py
from google.cloud import logging

client = logging.client.Client()
metric = logging.metric.Metric()
```

Daniel-Sanche: Yeah that first one looks good! 

What do you think we should do with the auto-generated API (currently in `logging_v2`)? I want to discourage users from using the auto-generated modules and use it mostly internally, but there may be some gaps in the hand-written code for now, and it may make sense to allow users to "break the glass". Maybe export them as `logging.gapic.MetricsServiceV2Client`? What do you think?

Also, an issue with `from google.cloud import logging` is that it collides with the `logging` standard library, and they are often [going to be used together](https://cloud.google.com/logging/docs/setup/python#connecting_the_library_to_python_logging). Right now we tell users to import as `import google.cloud.logging`. I wonder if it would make sense to rename the library to `google.cloud.cloudlogging` or something so there's no naming conflict

Daniel-Sanche: >  I can remove the generated clients from `__all__` and make them available through `logging_v2.gapic.MetricServiceV2Client`. This is similar to what you have to do to get to the transport layer in microgenerated clients.

Ok, that would be great!

> Ah I didn't realize that was a problem. This is definitely a good time to do any naming changes. The generator is flexible on namespacing, so it is really up to you.

Hmm ok, I'll discuss this with the team and get back to you. Let me know if you have strong opinions one way or the other, because I can definitely see pros and cons to each

busunkim96: Alright, I've removed the 3 generated clients from `logging_v2`/`logging`. PTAL.

```py
from google.cloud import logging_v2

logging_v2.gapic.logging_service_v2_client.LoggingServiceV2Client()
```

Daniel-Sanche: It looks good for the most part, but I noticed that some modules are still only accessible through `logging_v2` rather than `logging`. Like `google.cloud.logging_v2.sink` and `google.cloud.logging_v2.metric` . Shouldn't the alias be the same for both?

daniel-sanche (approved):

PR #92, opened 2020-11-11
feat!: use microgenerator

busunkim96:
The bulk of the changes are to the handwritten code (`gogle/cloud/logging_v2/*.py`). Changes are listed in `UPGRADING.md`.

* `_gapic` uses the microgenerated surface
* Add support for parent resources that are not `project` (`folder`, `billingAccount`, `organization`) where appropriate for log entries and sinks.
* Use google-style docstrings
* Optional params can only be passed as kwargs


busunkim96:
@Daniel-Sanche This is now ready for review.

UPGRADING.md
```diff
@@ -0,0 +1,335 @@
...
+
+This section describes changes in the GAPIC layer (produced by the generator) that previously lived in  `google.cloud.logging_v2.proto` / `google.cloud.logging_v2.gapic`.
+
+> **NOTE**: Most users are unlikely to have been using this layer directly.
+
+### Import path
+
+> **WARNING**: Breaking change
+
+The generated client is no longer exposed at `logging_v2`.
```
Daniel-Sanche, line 125: It might be good to mention why here. Maybe link to the API docs for the hand-written layer?

UPGRADING.md
```diff
@@ -0,0 +1,335 @@
...
+from google.cloud.logging_v2.types import LogSink
+
+logging_client = LoggingServiceV2Client()
+sink = LogSink()
+```
+### Method Calls
+
+> **WARNING**: Breaking change
+
+Methods expect request objects. We provide a script that will convert most common use cases.
```
Daniel-Sanche, line 146: This should only affect users using the gapic layer, right?

daniel-sanche (approved):
I looked over most of _gapic, and read the upgrading guide, and skimmed over some of the top-level files. It looks like as long as the tests are happy, LGTM. Let me know if you think there are any other files that need extra attention though

UPGRADING.md
busunkim96: Yep, I'll add another note to make that more clear.

google/cloud/logging_v2/proto/logging.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC.
```
software-dov, line 1: Shouldn't this stay as 2019?

PR #102, opened 2020-11-20
feat!: Add option to omit logging.StreamHandler during setup during setup

### LINKED ISSUE #38: Add option to omit `logging.StreamHandler` during setup
This issue is in regards to the `setup_logging` method on the `google.cloud.logging.Client`.

### Is your feature request related to a problem? Please describe.

The problem is that my logs are getting duplicated.

```py
import google.cloud.logging
client.setup_logging()
```

When I follow the simple example in [the docs](https://cloud.google.com/logging/docs/setup/python#connecting_the_library_to_python_logging) to connect my root logger to GCP logging, it sets up 2 handlers, a `logging.StreamHandler` and a `google.cloud.logging.handlers.CloudLoggingHandler`. This is often very reasonable. However, it is causing problems in my case because I also have [this set up](https://cloud.google.com/monitoring/kubernetes-engine/installing#migrating) to send ALL container logs to cloud logging.
The result is a duplication of my logs, one copy from the python root logger, and another copy getting picked up from stderr. (Note that the copy coming from stderr always has its level set to ERROR). This is a problem because the logs are much noisier than needed, causing them to be less useful.

#### Why not change the **Stackdriver Kubernetes Engine Monitoring** option in your cluster?

I currently have it set to **System and workload logging and monitoring**. I *could* set it to **System logging and monitoring only** instead, which will prevent *this pod's* logs from being duplicated. However, there are other, simpler pods that do not have cloud logging set up. I need all of their logs to be captured, not just the pods for which cloud logging is configured. However, once cloud logging **is** configured, I want it to go through the `CloudLoggingHandler` because it has advantages like setting the correct log level, and I don't want the logs to be duplicated.

### Describe the solution you'd like

My issue is with the implementation of [this method](https://github.com/googleapis/python-logging/blob/c6eb60179d674dfd5137d90d209094c9369b3581/google/cloud/logging/handlers/handlers.py#L152). I would like an option that prevents this line `logger.addHandler(logging.StreamHandler())` from happening. Perhaps a new argument like `client.setup_logging(stream_handler=False)`.

### Describe alternatives you've considered

This is my current workaround

```py
from google.cloud import logging as gcp_logging
from google.cloud.logging.handlers import CloudLoggingHandler, ContainerEngineHandler

logging_client = gcp_logging.Client(project=PROJECT_ID)

logging_client.setup_logging(log_level=level)
root_logger = logging.getLogger()
# use the GCP handler ONLY in order to prevent logs from getting written to STDERR
root_logger.handlers = [handler
                        for handler in root_logger.handlers
                        if isinstance(handler, (CloudLoggingHandler, ContainerEngineHandler))]

```

### Additional context

Here is a screenshot of the console showing my duplicated logs:
Screen Shot 2020-05-04 at 5 11 32 PM

ZeJ0hn:

Fixes #38 🦕

### DIFF
--- a/google/cloud/logging_v2/client.py
+++ b/google/cloud/logging_v2/client.py
@@ -361,7 +361,12 @@ class Client(ClientWithProject):
             return CloudLoggingHandler(self, **kw)
 
     def setup_logging(
-        self, *, log_level=logging.INFO, excluded_loggers=EXCLUDED_LOGGER_DEFAULTS, **kw
+        self,
+        *,
+        log_level=logging.INFO,
+        excluded_loggers=EXCLUDED_LOGGER_DEFAULTS,
+        stream_handler=True,
+        **kw,
     ):
         """Attach default Cloud Logging handler to the root logger.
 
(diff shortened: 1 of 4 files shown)


ZeJ0hn:
@googlebot I fixed it.

ZeJ0hn:
@googlebot I fixed it.

PR #105, opened 2020-11-23
tests: skip some system tests for mtls testing

arithmetic1728:
Skip some bigquery/storage/pubsub system test cases that don't work for mtls testing.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -14,6 +14,8 @@
 
 import datetime
 import logging
+import os
+import pytest
 import unittest
 
 from google.api_core.exceptions import BadGateway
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):
LGTM, but please let @Daniel-Sanche take a look as well.

Can you create an issue to track re-enabling these tests later?

daniel-sanche (commented):
Do you have any more context for these? The tests seem to be passing fine on kokoro, do you have a different test system somewhere that requires mTLS? Or is this for a future change to the kokoro tests?

arithmetic1728:
@Daniel-Sanche There is nothing wrong with the tests or kokoro. We use a special internal testing environment to test mTLS. This PR just adds a tag so the internal testing environment can skip the tests not suitable for mTLS.

arithmetic1728:
> LGTM, but please let @Daniel-Sanche take a look as well.
> 
> Can you create an issue to track re-enabling these tests later?

The tests are not skipped for kokoro. This "skip_for_mtls" tag is only for the internal mTLS testing environment.

There is no need to re-enable them (also not possible for those using storage/bigquery http clients), because the remaining tests are enough for mTLS testing purposes.

daniel-sanche (approved):
Ok sounds good, thanks!

PR #113, opened 2020-12-03
fix: Remove keyword only argument for RequestsMiddleware

### LINKED ISSUE #112: Keyword only args causes RequestMiddleware to fail on Django app load
#### Environment details

  - OS type and version: macOS version 11.0.1
  - Python version: Python 3.8.2
  - pip version: pip 20.2.4
  - `google-cloud-logging` version: 2.0.1 and 2.0.0

#### Steps to reproduce

  1. Add `google.cloud.logging_v2.handlers.middleware.RequestMiddleware` to a `MIDDLEWARE` list in django app settings
  2. Make an API request, e.g. during a test 

#### Stack trace
```
self = <django.test.client.ClientHandler object at 0x124a5f850>

    def load_middleware(self):
        """
        Populate middleware lists from settings.MIDDLEWARE.

        Must be called after the environment is fixed (see __call__ in subclasses).
        """
        self._view_middleware = []
        self._template_response_middleware = []
        self._exception_middleware = []

        handler = convert_exception_to_response(self._get_response)
        for middleware_path in reversed(settings.MIDDLEWARE):
            middleware = import_string(middleware_path)
            try:
>               mw_instance = middleware(handler)
E               TypeError: __init__() takes 1 positional argument but 2 were given
```

#### Extra Info
It looks like django [doesn't support keyword only arguments](https://github.com/django/django/blob/d8dfff2ab0edf7a1ca5255eccf45c447b2f9d57e/django/core/handlers/base.py#L58) when loading middleware (the link is pointing to master, but it's the same for old versions). If I locally do the following, I can subclass the `RequestMiddleware` and remove the keyword only args and the whole issue goes away

bradykieffer:
Remove keyword only arguments from request middleware. This causes
django to fail when attempting to load middleware. Django currently only
supports handlers being passed in as args.

Fixes #112 🦕

### DIFF
--- a/google/cloud/logging_v2/handlers/middleware/request.py
+++ b/google/cloud/logging_v2/handlers/middleware/request.py
@@ -42,7 +42,7 @@ except ImportError:  # pragma: NO COVER
 class RequestMiddleware(MiddlewareMixin):
     """Saves the request in thread local"""
 
-    def __init__(self, get_response=None):
+    def __init__(self, get_response):
         self.get_response = get_response
 
     def process_request(self, request):
--- a/google/cloud/logging_v2/handlers/middleware/request.py
+++ b/google/cloud/logging_v2/handlers/middleware/request.py
@@ -42,7 +42,7 @@ except ImportError:  # pragma: NO COVER
 class RequestMiddleware(MiddlewareMixin):
     """Saves the request in thread local"""
 
-    def __init__(self, *, get_response=None):
+    def __init__(self, get_response=None):
         self.get_response = get_response
 
     def process_request(self, request):
--- a/tests/unit/handlers/middleware/test_request.py
+++ b/tests/unit/handlers/middleware/test_request.py
@@ -54,6 +54,16 @@ class TestRequestMiddleware(DjangoBase):
         django_request = request._get_django_request()
         self.assertEqual(django_request, mock_request)
 
+    def test_can_instantiate_middleware_without_kwargs(self):
+        handler = mock.Mock()
+        middleware = self._make_one(handler)
+        self.assertEqual(middleware.get_response, handler)
+
+    def test_can_instantiate_middleware_with_kwargs(self):
+        handler = mock.Mock()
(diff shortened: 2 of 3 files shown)


bradykieffer:
@googlebot I signed it!

Daniel-Sanche:
Good catch, thanks! Do you think you could add a [unit test](https://github.com/googleapis/python-logging/blob/master/tests/unit/handlers/middleware/test_request.py) to catch this going forward?

daniel-sanche (approved):

bradykieffer:
Of course @Daniel-Sanche! I broke my rule about always pushing up a patch with a test, especially when there's a bug. I commented on #112:

>A colleague of mine pointed out that this issue is related to the 2.0.0 update. Looking at Django's docs for 3.1 and 2.2 it doesn't seem like get_response is an optional, we could default it to non-null to fix this issue.

Do you know if code gen will cause this to fail going forward? If so, I _think_ it's safe make `get_response` a required param

Daniel-Sanche:
>Of course @Daniel-Sanche! I broke my rule about always pushing up a patch with a test, especially when there's a bug. I commented on #112:

Perfect, thanks!

> A colleague of mine pointed out that this issue is related to the 2.0.0 update. Looking at Django's docs for 3.1 and 2.2 it doesn't seem like get_response is an optional, we could default it to non-null to fix this issue.

Yeah, I just took a look at the [django documentation](https://docs.djangoproject.com/en/3.1/topics/http/middleware/#init-get-response) and that seems to be the case, so that sounds like a good idea. 

> Do you know if code gen will cause this to fail going forward? If so, I think it's safe make get_response a required param

The code generator won't be an issue here, it should only impact `/proto`, `/services`, and `/types`. This file is pretty stand-alone

bradykieffer:
Awesome, I added tests + made `get_response` a required param

Daniel-Sanche:
Great, thanks for your contribution!

PR #117, opened 2020-12-08
fix: Add submodule imports for handlers to logging alias

### LINKED ISSUE #115: handlers is not accessible from`google.cloud.logging`
#### Environment details

  - OS type and version: Linux Fedora 33
  - Python version: 3.9.0
  - pip version: 20.2.2
  - `google-cloud-logging` version: 2.0.1

#### Steps to reproduce

  1. install google-cloud-logging
  2. open a python session in your terminal

#### Code example

handlers is not accessible from`google.cloud.logging` :

```
>>> from google.cloud.logging.handlers import CloudLoggingHandler
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'google.cloud.logging.handlers'
```

Whereas it is from `google.cloud.logging_v2` :

```
>>> from google.cloud.logging_v2.handlers import CloudLoggingHandler
>>> 
```
```

### PR DESCRIPTION
**Pull request by daniel-sanche**
We give users two packages: `logging_v2` for the current version of the library, and `logging` as the alias. 

We found the alias was only exporting paths at the library root, which was not what users expected. This meant they could not run 
```from google.cloud.logging.handlers import CloudLoggingHandler```.

Instead they had to run 
```
from google.cloud.logging import handlers
handlers.CloudLoggingHandler
```

This PR adds more `__init__.py` files to advertise sub-modules for the handlers. I avoided doing the same for the gapic directories (`/proto`, `/services`, `/types`) because we want to discourage end-users from using auto-generated code

Fixes https://github.com/googleapis/python-logging/issues/115 🦕

### DIFF
--- a/google/cloud/logging/handlers/__init__.py
+++ b/google/cloud/logging/handlers/__init__.py
@@ -1,0 +1,27 @@
+# Copyright 2016 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Python :mod:`logging` handlers for Google Cloud Logging."""
+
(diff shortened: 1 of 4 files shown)


Daniel-Sanche:
@busunkim96 let me know if you know a more elegant way to do this

busunkim96 (commented):
This looks good to me. Could you add some unit tests to https://github.com/googleapis/python-logging/blob/master/tests/unit/test_logging_shim.py to make sure this stays in sync with `logging_v2.handlers`?

busunkim96 (approved):

Issue #110, opened 2020-12-03
Log statements are not grouped to parent request in App Engine Standard P3


MeLight:
#### Environment details

  - OS type and version: App Engine Python 3 Standard
#### Steps to reproduce

  1. Add couple of logging statements 
  2. Go to Logs Explorer on your cloud console

#### Code example

```python
import google.cloud.logging
import logging

client = google.cloud.logging.Client()

client.get_default_handler()
client.setup_logging()

logging.info("Testing lines")
logging.info("Again!")
```

Result:
![image](https://user-images.githubusercontent.com/704484/100991402-9f8c4c00-355b-11eb-84f5-a3ed45152d59.png)
As you can see every statement is on it's own line, as opposed to what's expected - the statements should be grouped under one log like in App Engine Standard Python2
![image](https://user-images.githubusercontent.com/704484/100995719-d1ec7800-3560-11eb-967a-2dbaba2bd681.png)

How to group the statements to the parent request?

snarfed:
I'm also seeing this, no matter what I try. Looks like it's the default with google-cloud-logging 2.0 on App Engine Standard: the built in Python root logger sends to Cloud Logging, with resource type `gae_app` and log name `projects/brid-gy/logs/app` - good! - but those log lines aren't attached and rendered with the `appengine.googleapis.com/request_log` log entry.

I first noticed this change when I migrated to App Engine Python 3 a while back. I managed to fix it for google-cloud-logging 1.x with the code below, [which I found here](https://stackoverflow.com/a/58296028/186123):

```
import google.cloud.logging
from google.cloud.logging_v2.handlers import AppEngineHandler, setup_logging
logging_client = google.cloud.logging.Client()
setup_logging(AppEngineHandler(logging_client, name='stdout'))
```

... but this no longer works with google-cloud-logging 2.0. I've tried a bunch of variants, including no manual setup at all (ie relying on the [default integration on App Engine](https://cloud.google.com/logging/docs/setup/python#run-gae)), but no luck so far.

Daniel-Sanche:
Thanks the the report. We're working on improving the cross-platform tests to make sure everything's consistent across GCP platforms, and this bug should be fixed by that effort. I'll try to bump of the priority of this one though, and see if I can fix it a bit earlier

What framework are you using for your web server to handle these requests?

snarfed:
Thank you @Daniel-Sanche! Sounds good. I'll also try downgrading to 1.x in the meantime to see if that fixes it.

[...]

PR #120, opened 2020-12-10
feat: support http_request field

### LINKED ISSUE #72: Improve Flask integration
flask HTTP requests should be captured as part of the LogEntry sent to GCP. See other libraries like https://pypi.org/project/flask-gcp-log-groups/

daniel-sanche:
This PR adds support for the [`http_request`](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest) field that's part of the Cloud Logging LogEntry spec. Details about the request (method, url, payload size, user agent, remote ip, referrer) will be attached as metadata to each log when possible. These fields will be inferred from Flask and Django environments, just as the `trace_id` currently is. 

I plan to allow users to set this information manually as well for users using unsupported libraries, but that will come in a separate PR

Currently only fully supported on AppEngine

This fixes https://github.com/googleapis/python-logging/issues/72, partially addresses https://github.com/googleapis/python-logging/issues/110, and lays some groundwork for later improvements

### DIFF
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -115,7 +115,7 @@ def get_request_data_from_django():
 
 def get_request_data():
     """Helper to get trace_id and http_request data from supported web
-    frameworks.
+    frameworks (currently supported: Flask and Django).
 
     Returns:
         str: TraceID in HTTP request headers.
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -62,10 +62,11 @@ def get_request_data_from_flask():
         return None, None
 
     # build http_request
+    length = flask.request.content_length
     http_request = {
         "request_method": flask.request.method,
         "request_url": flask.request.url,
-        "request_size": str(flask.request.content_length),
+        "request_size": str(length) if length else None,
         "user_agent": flask.request.user_agent.string,
         "remote_ip": flask.request.remote_addr,
         "referer": flask.request.referrer,
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -58,6 +58,8 @@ def get_request_data_from_flask():
         str: TraceID in HTTP request headers.
         HttpRequest: data about the associated http request.
     """
+    if flask is None or not flask.request:
+        return None, None
 
     # build http_request
     http_request = HttpRequest(
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -30,6 +30,7 @@ _DJANGO_USERAGENT_HEADER = "HTTP_USER_AGENT"
 _DJANGO_REMOTE_ADDR_HEADER = "REMOTE_ADDR"
 _DJANGO_REFERER_HEADER = "HTTP_REFERER"
 _FLASK_TRACE_HEADER = "X_CLOUD_TRACE_CONTEXT"
+_PROTOCOL_HEADER = "SERVER_PROTOCOL"
 
 
 def format_stackdriver_json(record, message):
@@ -68,6 +69,7 @@ def get_request_data_from_flask():
         "user_agent": flask.request.user_agent.string,
         "remote_ip": flask.request.remote_addr,
         "referer": flask.request.referrer,
+        "protocol": request.environ.get(_PROTOCOL_HEADER)
     }
 
     # find trace id
@@ -99,6 +101,7 @@ def get_request_data_from_django():
         "user_agent": request.META.get(_DJANGO_USERAGENT_HEADER),
         "remote_ip": request.META.get(_DJANGO_REMOTE_ADDR_HEADER),
         "referer": request.META.get(_DJANGO_REFERER_HEADER),
+        "protocol": request.META.get(_PROTOCOL_HEADER),
     }
 
     # find trace id
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -63,8 +63,8 @@ def get_request_data_from_flask():
     # build http_request
     http_request = {
         'request_method': flask.request.method,
-        'request_url': flask.request.host_url,
-        'request_size': flask.request.content_length,
+        'request_url': flask.request.url,
+        'request_size': str(flask.request.content_length),
         'user_agent': flask.request.user_agent.string,
         'remote_ip': flask.request.remote_addr,
         'referer': flask.request.referrer,
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -93,7 +93,7 @@ def get_request_data_from_django():
     # build http_request
     http_request = {
         'request_method': request.method,
-        'request_url': request.get_full_path(),
+        'request_url': request.build_absolute_uri(),
         'request_size': request.META.get(_DJANGO_LENGTH_HEADER),
         'user_agent': request.META.get(_DJANGO_USERAGENT_HEADER),
         'remote_ip': request.META.get(_DJANGO_REMOTE_ADDR_HEADER),
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -51,10 +51,11 @@ def format_stackdriver_json(record, message):
 
 
 def get_request_data_from_flask():
-    """Get trace_id from flask request headers.
+    """Get trace_id and http_request data from flask request headers.
 
     Returns:
         str: TraceID in HTTP request headers.
+        dict: data about the associated http request.
     """
     if flask is None or not flask.request:
         return None, None
@@ -78,10 +79,11 @@ def get_request_data_from_flask():
     return trace_id, http_request
 
 def get_request_data_from_django():
-    """Get trace_id from django request headers.
+    """Get trace_id and http_request data from django request headers.
 
     Returns:
         str: TraceID in HTTP request headers.
+        dict: data about the associated http request.
     """
     request = _get_django_request()
 
@@ -109,10 +111,12 @@ def get_request_data_from_django():
 
 
 def get_request_data():
-    """Helper to get trace_id from web application request header.
+    """Helper to get trace_id and http_request data from supported web
+    frameworks.
 
     Returns:
         str: TraceID in HTTP request headers.
+        dict: data about the associated http request.
     """
     checkers = (
         get_request_data_from_django,
@@ -121,7 +125,7 @@ def get_request_data():
 
     for checker in checkers:
         trace_id, http_request = checker()
-        if trace_id is not None:
+        if http_request is not None:
             return trace_id, http_request
 
-    return None
+    return None, None
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -64,6 +64,24 @@ def get_trace_id_from_flask():
 
     return trace_id
 
+def get_http_request_from_flask():
+    """Get trace_id from flask request headers.
+
+    Returns:
+        str: TraceID in HTTP request headers.
+    """
+    if flask is None or not flask.request:
+        return None
+
+    obj = {'request_method':flask.request.method,
+            'request_url': flask.request.host_url,
+            'request_size': flask.request.content_length,
+            'user_agent': flask.request.user_agent.string,
+            'remote_ip': flask.request.remote_addr,
+            'referer': flask.request.referrer,
+            }
+
+    return obj
 
 def get_trace_id_from_django():
     """Get trace_id from django request headers.
@@ -100,5 +118,23 @@ def get_trace_id():
         trace_id = checker()
         if trace_id is not None:
             return trace_id
+
+    return None
+
+
+def get_http_request_data():
+    """Helper to get trace_id from web application request header.
+
+    Returns:
+        str: TraceID in HTTP request headers.
+    """
+    checkers = (
+        get_http_request_from_flask,
+    )
+
+    for checker in checkers:
+        obj = checker()
+        if obj  is not None:
+            return obj
 
     return None
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -52,11 +52,12 @@ def format_stackdriver_json(record, message):
 
 
 def get_request_data_from_flask():
-    """Get trace_id and http_request data from flask request headers.
+    """Get http_request and trace data from flask request headers.
 
     Returns:
-        str: TraceID in HTTP request headers.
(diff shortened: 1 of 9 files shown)


google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -46,59 +50,82 @@ def format_stackdriver_json(record, message):
...
     Returns:
         str: TraceID in HTTP request headers.
+        dict: data about the associated http request.
     """
     if flask is None or not flask.request:
-        return None
+        return None, None
+
+    # build http_request
+    http_request = {
```
nicoleczhu, line 64: You might be able to get or be able to infer other fields as mentioned by the HttpRequest [spec here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest), e.g. `protocol` 

Caveat: Im not familiar with what's exposed in request types in django/flask

google/cloud/logging_v2/handlers/app_engine.py
```diff
@@ -114,11 +114,14 @@ def emit(self, record):
...
         message = super(AppEngineHandler, self).format(record)
         gae_labels = self.get_gae_labels()
-        trace_id = (
-            "projects/%s/traces/%s" % (self.project_id, gae_labels[_TRACE_ID_LABEL])
-            if _TRACE_ID_LABEL in gae_labels
-            else None
-        )
+        http_request, trace_id = get_request_data()
+        if trace_id is not None:
+            trace_id = f"projects/{self.project_id}/{trace_id}"
```
nicoleczhu, line 119: Is it `f"projects/{self.project_id}/{trace_id}"`
or is it f"projects/{self.project_id}/traces/{trace_id}"

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -46,59 +50,82 @@ def format_stackdriver_json(record, message):
...
+        "user_agent": flask.request.user_agent.string,
+        "remote_ip": flask.request.remote_addr,
+        "referer": flask.request.referrer,
+    }
 
+    # find trace id
+    trace_id = None
     header = flask.request.headers.get(_FLASK_TRACE_HEADER)
+    if header:
+        trace_id = header.split("/", 1)[0]
```
nicoleczhu, line 77: Future edge case to take care of: 

Given user input is `X-Cloud-Trace-Context: TRACE_ID/SPAN_ID;o=TRACE_TRUE` where any of the values can be nil or 0. This logic to split by / and grab the first value might give you `span id` instead. 

Seems super niche edge case, but I actually got a similar ticket for this in Go.

google/cloud/logging_v2/handlers/app_engine.py
Daniel-Sanche: Good catch! I'm not too familiar with tracing yet so I don't know what it *should* be, but I meant to copy what was being done previously. I'll have to add a test to check for this

google/cloud/logging_v2/handlers/_helpers.py
Daniel-Sanche: Hmm I'm not sure I understand. Can you have a span without a trace? Do you have a ticket or doc I can look over?

Either way, I'm not a huge fan of this parsing logic. I'm hoping to fix it up when I add support for spans soon

simonz130: It shouldn't be possible to have a trace as null but span as non-null. Curious how this was a case in Go?

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -46,59 +50,82 @@ def format_stackdriver_json(record, message):
...
+        return None, None
+
+    # build http_request
+    http_request = {
+        "request_method": request.method,
+        "request_url": request.build_absolute_uri(),
+        "request_size": request.META.get(_DJANGO_LENGTH_HEADER),
+        "user_agent": request.META.get(_DJANGO_USERAGENT_HEADER),
+        "remote_ip": request.META.get(_DJANGO_REMOTE_ADDR_HEADER),
+        "referer": request.META.get(_DJANGO_REFERER_HEADER),
```
simonz130, line 101: can some of these not be set? What happens if there is no value for a key in request.META?

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -46,59 +50,82 @@ def format_stackdriver_json(record, message):
...
 
-    return trace_id
+    return http_request, trace_id
 
 
-def get_trace_id():
-    """Helper to get trace_id from web application request header.
+def get_request_data():
+    """Helper to get trace_id and http_request data from supported web
+    frameworks.
```
simonz130, line 115: probably worth clarifying which frameworks are currently supported in comment

tests/unit/handlers/test__helpers.py
```diff
@@ -89,7 +140,11 @@ def test_no_context_header(self):
 
         middleware = request.RequestMiddleware(None)
         middleware.process_request(django_request)
-        trace_id = self._call_fut()
+        http_request, trace_id = self._call_fut()
+        self.assertEqual(http_request["request_method"], "GET")
+        self.assertEqual(set(http_request.keys()), set(_HTTP_REQUEST_FIELDS))
+        for field in _HTTP_REQUEST_FIELDS:
+            self.assertTrue(field in http_request)
```
simonz130, line 147: not familiar with the test framework in python - is this method overload printing the message to add context of what is expected vs actual?

E.g.
instead of:
```
expected: "abc", actual: "xyz"
```
it should be 
``` 
expected:"abc", actual:"xyz" **for <<field>>**
```

tests/unit/handlers/test__helpers.py
```diff
@@ -106,61 +161,122 @@ def test_valid_context_header(self):
...
-        trace_id = self._call_fut()
+        http_request, trace_id = self._call_fut()
 
         self.assertEqual(trace_id, expected_trace_id)
+        self.assertEqual(http_request["request_method"], "GET")
+        self.assertEqual(set(http_request.keys()), set(_HTTP_REQUEST_FIELDS))
+        for field in _HTTP_REQUEST_FIELDS:
+            self.assertTrue(field in http_request)
 
+    def test_http_data(self):
```
simonz130, line 172: please name the test so it's more concrete in name what is being tested. What exactly in http data?

tests/unit/handlers/test__helpers.py
```diff
@@ -106,61 +161,122 @@ def test_valid_context_header(self):
...
 
         django_mock.assert_called_once_with()
         flask_mock.assert_not_called()
 
-    def test_missing(self):
-        django_mock, flask_mock, trace_id = self._helper(None, None)
-        self.assertIsNone(trace_id)
+    def test_missing_http_request(self):
+        flask_expected = (None, _FLASK_TRACE_ID)
+        django_expected = (None, _DJANGO_TRACE_ID)
```
simonz130, line 255: perhaps, worth splitting this into two separate tests - for django and for flask

tests/unit/handlers/test_app_engine.py
```diff
@@ -87,36 +87,54 @@ def test_constructor_w_gae_flex_env(self):
...
-            handler.transport.send_called_with,
-            (record, message, gae_resource, gae_labels, trace),
+        expected_http_request = {"request_url": "test"}
+        trace_id = "trace-test"
+        expected_trace_id = f"projects/{self.PROJECT}/traces/{trace_id}"
+        get_request_patch = mock.patch(
+            "google.cloud.logging_v2.handlers.app_engine.get_request_data",
+            return_value=({"request_url": "test"}, "trace-test"),
         )
+        with get_request_patch:
```
simonz130, line 97: add comments to explain that you test attached http request properties

google/cloud/logging_v2/handlers/_helpers.py
Daniel-Sanche: Some of them I can't include, because the request isn't finished (responseSize, status, etc). Others I'm not quite sure about, like the cache fields. Do you know if that should be in the headers somewhere?

For some reason Flask didn't make the request's protocol as easy to find as the others, but it should be right there in the header, so I'll take another look there

google/cloud/logging_v2/handlers/_helpers.py
Daniel-Sanche: I'm believe they default to None, but good point, I'll add a test to be sure

tests/unit/handlers/test__helpers.py
Daniel-Sanche: Good point, I should have added a message for the test here. 

But I actually realized this assertion is redundant with the one above it, and it already gives a clear error. So I pulled this one out

tests/unit/handlers/test__helpers.py
Daniel-Sanche: This test specifically is focused on the integration function that grabs from either django or flask, the django and flask components are mocked out here

google/cloud/logging_v2/handlers/_helpers.py
tseaver: `{}.get("nonesuch-key")` does indeed return `None`.

google/cloud/logging_v2/handlers/_helpers.py
nicoleczhu: The edge cases I'm referring to here is just that the `X-Cloud-Trace-Context` field is provided by the user. 

So the user could manually write anything here. e.g. `'/hello/world'` and we'd grep the first token as trace_id. Adding better regex later on could help.

[This](https://github.com/googleapis/google-cloud-go/issues/1590) was the original Go issue.

nicoleczhu: Where we tell users they can manually write this header: https://cloud.google.com/trace/docs/setup#force-trace

Daniel-Sanche: Hmm good to keep in mind, yeah it might be helpful to throw an exception or something if the header doesn't match the expected format

I'll probably fix that in a later PR when I add support for spans though. That's when proper parsing will be more important. For now, this is the logic that was already in place

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -46,59 +51,83 @@ def format_stackdriver_json(record, message):
...
 
 
-def get_trace_id_from_django():
-    """Get trace_id from django request headers.
+def get_request_data_from_django():
+    """Get trace_id and http_request data from django request headers.
 
     Returns:
         str: TraceID in HTTP request headers.
+        HttpRequest: data about the associated http request.
```
busunkim96, line 89: Can you adjust the 'Returns' to show that this returns a Tuple? See [example](https://github.com/googleapis/google-auth-library-python/blob/0323cf390b16e8483660ac88775e8ea4e7f7702d/google/auth/_default.py#L315-L319)

google/cloud/logging_v2/handlers/transports/background_thread.py
```diff
@@ -222,31 +222,21 @@ def _main_thread_terminated(self):
...
             record (logging.LogRecord): Python log record that the handler was called with.
             message (str): The message from the ``LogRecord`` after being
                         formatted by the associated log formatters.
-            resource (Optional[google.cloud.logging_v2.resource.Resource]):
-                Monitored resource of the entry
-            labels (Optional[dict]): Mapping of labels for the entry.
-            trace (Optional[str]): TraceID to apply to the logging entry.
-            span_id (Optional[str]): Span_id within the trace for the log entry.
-                Specify the trace parameter if span_id is set.
+            **kwargs: Additional optional arguments for the logger
```
busunkim96, line 232:
```suggestion
            kwargs: Additional optional arguments for the logger
```

google/cloud/logging_v2/handlers/transports/background_thread.py
```diff
@@ -291,30 +281,16 @@ def __init__(
...
             record (logging.LogRecord): Python log record that the handler was called with.
             message (str): The message from the ``LogRecord`` after being
                 formatted by the associated log formatters.
-            resource (Optional[google.cloud.logging_v2.resource.Resource]):
-                Monitored resource of the entry.
-            labels (Optional[dict]): Mapping of labels for the entry.
-            trace (Optional[str]): TraceID to apply to the logging entry.
-            span_id (Optional[str]): span_id within the trace for the log entry.
-                Specify the trace parameter if span_id is set.
+            **kwargs: Additional optional arguments for the logger
```
busunkim96, line 291:
```suggestion
            kwargs: Additional optional arguments for the logger
```

google/cloud/logging_v2/handlers/transports/base.py
```diff
@@ -22,18 +22,14 @@ class Transport(object):
...
         """Transport send to be implemented by subclasses.
 
         Args:
             record (logging.LogRecord): Python log record that the handler was called with.
             message (str): The message from the ``LogRecord`` after being
                 formatted by the associated log formatters.
-            resource (Optional[google.cloud.logging_v2.resource.Resource]):
-                 Monitored resource of the entry.
-            labels (Optional[dict]): Mapping of labels for the entry.
+            **kwargs: Additional optional arguments for the logger
```
busunkim96, line 32:
```suggestion
            kwargs: Additional optional arguments for the logger
```

google/cloud/logging_v2/handlers/transports/sync.py
```diff
@@ -30,26 +30,17 @@ class SyncTransport(Transport):
...
 
         Args:
             record (logging.LogRecord):
                 Python log record that the handler was called with.
             message (str): The message from the ``LogRecord`` after being
                 formatted by the associated log formatters.
-            resource (Optional[~logging_v2.resource.Resource]):
-                 Monitored resource of the entry.
-            labels (Optional[dict]): Mapping of labels for the entry.
+            **kwargs: Additional optional arguments for the logger
```
busunkim96, line 41:
```suggestion
            kwargs: Additional optional arguments for the logger
```

busunkim96 (approved):

google/cloud/logging_v2/handlers/app_engine.py
```diff
@@ -114,11 +114,14 @@ def emit(self, record):
...
+        if trace_id is not None:
+            trace_id = f"projects/{self.project_id}/traces/{trace_id}"
         self.transport.send(
-            record, message, resource=self.resource, labels=gae_labels, trace=trace_id
+            record,
+            message,
+            resource=self.resource,
+            labels=gae_labels,
+            trace=trace_id,
+            http_request=http_request,
```
busunkim96, line 126: I see you're using `http_pb2_request` does that work properly with the `LogEntry`?https://github.com/googleapis/python-logging/blob/6349b899811cbb16f5548df0b77564b46666c4e7/google/cloud/logging_v2/entries.py#L173

Daniel-Sanche: I believe it should. This is the type used by the underlying LogEntry proto, so I assumed it would be preferred: https://github.com/googleapis/python-logging/blob/ff90fd2fb54c612fe6ab29708a2d5d984f60dea7/google/cloud/logging_v2/types/log_entry.py#L192

But I'm still getting used to working with gapic code. Do you prefer to just use dictionaries for these kinds of fields?

PR #129, opened 2020-12-11
feat: allow modifying LogEntry data using extra argument

### LINKED ISSUE #110: Log statements are not grouped to parent request in App Engine Standard P3

daniel-sanche:
Our library allows users to use the standard Python logging library to send logs to Stackdriver. So far, this method has had limitations, because users couldn't customize any log data associated with the message. This PR allows users to send in metadata fields using the standard [`extra` dictionary parameter](https://docs.python.org/3/library/logging.html#logging.Logger.debug), allowing users to override fields like `trace`, `span_id`, `http_request`, and `labels`.

```
logging.info('log test', extra={'trace':'12345''})
```

So far, this change will only work on AppEngine and the using the base CloudLoggingHandler, but I'll slowly roll out support to more environments over time

Fixes https://github.com/googleapis/python-logging/issues/110 and lays the groundwork for later changes


freelerobot (approved):
Food for thought: 

1. Will users be able to also modify additional fields like timestamp, insertID, etc permitted by the spec
2. What about letting users define `extras` at the log client level, e.g. `client.setup_logging(extras=...)`
 3. Will you/do you need to communicate in documentation to users that this is only enabled for AppEngine

Daniel-Sanche:
That's all really good feedback:
1) This is probably one of the things we'd want to keep consistent across languages. Are the other libraries permissive about overwriting those fields?
2) That's a great suggestion. I might consider making another PR for that before cutting this release.
3) My plan was to keep this minimally documented for now (and just point customers to it when needed), and document it better when it works on all environments. But if you have other ideas or suggestions I'm interested in hearing them

PR #151, opened 2021-01-11
fix: add InternalServerError to list of expected errors

daniel-sanche:
We are currently seeing a failure on the continuous CI tests for this repo. It looks like an internal server error is showing up on one of the list_logs calls. This fix adds that exception to the list of exceptions to retry

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -20,6 +20,7 @@ import unittest
 
 from google.api_core.exceptions import BadGateway
 from google.api_core.exceptions import Conflict
+from google.api_core.exceptions import InternalServerError
 from google.api_core.exceptions import NotFound
 from google.api_core.exceptions import TooManyRequests
 from google.api_core.exceptions import ResourceExhausted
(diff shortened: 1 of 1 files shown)


tests/system/test_system.py
```diff
@@ -67,7 +68,9 @@ def _list_entries(logger):
     :returns: List of all entries consumed.
     """
     inner = RetryResult(_has_entries, max_tries=9)(_consume_entries)
-    outer = RetryErrors((ServiceUnavailable, ResourceExhausted), max_tries=9)(inner)
+    outer = RetryErrors(
```
busunkim96, line 71: I think this is OK for the purposes of tests. For any changes to automatic retries in the client make sure to reference https://google.aip.dev/194 (it has guidance on error codes that should/shouldn't be retried).

busunkim96 (approved):

tests/system/test_system.py
Daniel-Sanche: Good to know, thanks!

PR #161, opened 2021-01-25
test: reduced time filter length in system tests

daniel-sanche:
CI system tests have been failing for a week now. It looks like a rogue service has been flooding the test project with logs, which made our tests timeout

We fixed the problem on the project side, but I'm also introducing a change here to only query the last 10 minutes of logs. This should avoid similar problems from breaking our tests in the future

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -67,15 +67,13 @@ def _list_entries(logger, max_tries=5):
                 raise RuntimeError('no results found')
             else:
                 return entries
-        except (ServiceUnavailable, ResourceExhausted, InternalServerError,
-                RuntimeError) as e:
-            print(f'Error: {e}')
+        except:
             time.sleep(delay)
             try_num += 1
             delay *= 2
             if try_num >= max_tries:
                 # finished retries. Raise error again
-                raise e
+                raise
     return None
 
 
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -60,9 +60,11 @@ def _list_entries(logger, max_tries=5):
     delay = 1
     try_num = 0
     latest_error = None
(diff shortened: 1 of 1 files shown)


tests/system/test_system.py
```diff
@@ -41,19 +43,21 @@
...
+def _consume_entries(logger):
+    """Consume all recent log entries from logger iterator.
     :type logger: :class:`~google.cloud.logging.logger.Logger`
     :param logger: A Logger containing entries.
-
     :rtype: list
     :returns: List of all entries consumed.
     """
-    return list(logger.list_entries())
+    return list(logger.list_entries(filter_=_time_filter))
```
simonz130, line 60: nit: this looks really ugly. Is it a python convention to name variables with underscores as both the prefix and suffix? At least, maybe worth having a space before and after "=" for clarity.

simonz130 (approved):

busunkim96 (approved):

busunkim96:
Anything labeled as a `fix`, `feat`, or `doc` gets rolled up to the changelog by release-please.  I'd recommend using `test: ` for this PR.

tests/system/test_system.py
Daniel-Sanche: `filter_` is used because `fitler` is a keyword (I think it's part of the code generation process). Prefix underscores are more standard for internal variables.

Unfortunately we can't add spaces because it wouldn't be accepted by the linter. I think since this is all test code it should be fine though

Daniel-Sanche:
> Anything labeled as a `fix`, `feat`, or `doc` gets rolled up to the changelog by release-please. I'd recommend using `test: ` for this PR.

Good point. Switched to `test:`

PR #178, opened 2021-02-09
tests: update environment test script path

daniel-sanche:
The first kokoro run for the appengine standard environment test failed, saying it couldn't find the environment_tests.sh script. I updated the path to prefix with github, as that's the pattern I see in the samples tests. We'll see if that fixes the issue


busunkim96 (approved):
This seems to match the other `TRAMPOLINE_BUILD_FILE` paths. https://github.com/googleapis/python-logging/blob/69963b57054169e92ec05a1ec034c429c971f821/.kokoro/continuous/common.cfg#L26

PR #193, opened 2021-02-23
feat: Add json setting to allow unicodes to show in log instead of ascii ch…

ysde:
…aracter (for example, show chinese word '哈' in log instead of '\u8bed')

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #<192> 🦕

### DIFF
--- a/tests/unit/handlers/test_container_engine.py
+++ b/tests/unit/handlers/test_container_engine.py
@@ -40,7 +40,7 @@ class TestContainerEngineHandler(unittest.TestCase):
 
         handler = self._make_one()
         logname = "loggername"
-        message = "hello world"
+        message = "hello world，嗨 世界"
         record = logging.LogRecord(
             logname, logging.INFO, None, None, message, None, None
         )
@@ -53,4 +53,4 @@ class TestContainerEngineHandler(unittest.TestCase):
         }
         payload = handler.format(record)
 
-        self.assertEqual(payload, json.dumps(expected_payload))
+        self.assertEqual(payload, json.dumps(expected_payload, ensure_ascii=False))
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -33,7 +33,7 @@ _FLASK_TRACE_HEADER = "X_CLOUD_TRACE_CONTEXT"
 _PROTOCOL_HEADER = "SERVER_PROTOCOL"
 
 
-def format_stackdriver_json(record, message):
+def format_stackdriver_json(record, message, json_ensure_ascii):
     """Helper to format a LogRecord in in Stackdriver fluentd format.
 
     Returns:
@@ -48,7 +48,7 @@ def format_stackdriver_json(record, message):
         "severity": record.levelname,
     }
 
-    return json.dumps(payload)
+    return json.dumps(payload, ensure_ascii=json_ensure_ascii)
 
 
 def get_request_data_from_flask():
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -33,7 +33,7 @@ _FLASK_TRACE_HEADER = "X_CLOUD_TRACE_CONTEXT"
 _PROTOCOL_HEADER = "SERVER_PROTOCOL"
 
 
-def format_stackdriver_json(record, message, json_ensure_ascii):
+def format_stackdriver_json(record, message):
     """Helper to format a LogRecord in in Stackdriver fluentd format.
 
     Returns:
@@ -48,7 +48,7 @@ def format_stackdriver_json(record, message, json_ensure_ascii):
         "severity": record.levelname,
     }
 
-    return json.dumps(payload, ensure_ascii=json_ensure_ascii)
+    return json.dumps(payload, ensure_ascii=False)
 
 
 def get_request_data_from_flask():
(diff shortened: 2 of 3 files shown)


Daniel-Sanche:
Thanks for the contribution. I agree with the root problem (Unicode characters should be kept in the final logs), but I'm not sure if this is the best way to go about it:
- We should try to fix this in a way that applies to all environments, instead of just the Container Engine handler.
- It seems like unicode should be the default behavior, instead of an option that needs to be turned on.
- Before merging this, we'd want to make sure we put a bunch of tests in place (environment, system, and unit).

If you're interested in making those changes to this PR, that would be great. Otherwise, I can open a new issue and make sure it's addressed in a future release (this seems like it would be high priority on my priorities, so likely next release. But I may have to do some more testing to make sure)

ysde:
Hi @Daniel-Sanche 

Sure I'd like to make those changes to this PR. Do you have any suggestions where or which files should I started it?
And what should I do to pass all the checks ? 

Thank you very much

Daniel-Sanche:
The first step would be to add a new test to the new [environment test submodule](https://github.com/googleapis/python-logging/tree/master/tests) to see which GCP platforms are impacted. I [did that part myself here](https://github.com/googleapis/python-logging/pull/197) (you can take over the submodule, branch, or copy/paste the code)

It looks like only GKE is having issues there, which makes it simple (possibly cloud run too, but I think that's flakiness in the new tests - I'm looking into it). You can probably keep the `json.dumps(payload, ensure_ascii=False)` part of your code to fix it, but I don't think we need the other parts (I can't think of any situations where it would be useful to turn it off, unless you disagree)

After you have those fixes in and the environment test is passing, you should add similar checks to the [unit and system tests](https://github.com/googleapis/python-logging/tree/master/tests).

Our test system is in the process of being upgraded, so it may be a bit confusing/flaky. Feel free to reach out or re-assign the work back to me if you get stuck with the tests!

Daniel-Sanche:
Hey, any updates? If you're finding it hard to navigate the tests, I'm happy to take this over to get it in the next release. Let me know

ysde:
Hi @Daniel-Sanche 

Sorry for the late response, busy at working recently. :(

I'm not sure if there is some circumstances need to set `ensure_ascii=True`, so I was trying to keep the same behavior of the program, and think maybe add a `json_ensure_ascii` setting, default is true is good. 

I have little experience of using submodule and not sure how to take over [environment test submodule](https://github.com/googleapis/python-logging/tree/master/tests).

I've modified `test_container_engine.py`, but have `ModuleNotFoundError: No module named 'google.cloud'` when I try to execute `pytest -s  test_container_engine.py`

Can you give me this weekend to figure out how to finish it?

If it's releasing soon, you can take it over, no worries. : )

Thank you for the help and reviews.

Daniel-Sanche:
Ok, to make things simple you can ignore the environment test submodule, I can handle adding a test there after you're done.

For the unit tests, you should be able to run them through nox, which will simplify dependency issues. Information can be found in the repo's [CONTRIBUTING.md file](https://github.com/googleapis/python-logging/blob/master/CONTRIBUTING.rst#using-nox).

Let me know if you have other questions!

ysde:
@Daniel-Sanche 

Thanks! `nox` is so cool! I passed `nox -s --unit-3.6`

Do I have to install all python 3.6, 3.7, 3.8 to pass the tests?

I run `nox -s blacken`, `nox -s lint`, it will show messages like the following
`Session blacken skipped: Python interpreter 3.6 not found.`

And how to make the required checks pass?
```
Kokoro
Samples - Lint
Samples - Python 3.6
Samples - Python 3.7
Samples - Python 3.8
```
Thank you

tests/unit/handlers/test_container_engine.py
```diff
@@ -53,4 +53,4 @@ def test_format(self):
         }
         payload = handler.format(record)
 
-        self.assertEqual(payload, json.dumps(expected_payload))
+        self.assertEqual(payload, json.dumps(expected_payload, ensure_ascii=False))
```
Daniel-Sanche, line 56: I think this would be better as a separate test, rather than overriding the old `test_format`. You should duplicate this test and call it `test_format_unicode` or something

Also, it would be better to do an `self.assertEqual(payload['message'], message)`. That's a more direct test than relying on `json.dumps` to do what we expect

Daniel-Sanche:
> And how to make the required checks pass?


It looks like the CI tests are now passing. Are you still having issues?

tests/unit/handlers/test_container_engine.py
ysde: payload after `handler.format(record)` will be a `String`

How about like this?

```
expected_payload = json.dumps({
            "message": message,
            "timestamp": {"seconds": 5, "nanos": int(0.03 * 1e9)},
            "thread": record.thread,
            "severity": record.levelname,
        }, ensure_ascii=False)

        payload = handler.format(record)

        self.assertEqual(payload, expected_payload)
```

Thank you

ysde:
> > And how to make the required checks pass?
> 
> It looks like the CI tests are now passing. Are you still having issues?

Nope, CI's just not triggered that time, thank you !  :)

tests/unit/handlers/test_container_engine.py
Daniel-Sanche: Hmm good point. I'd probably prefer to check for a substring then: `self.assertTrue(message in payload)`

But if that doesn't work for some reason, I think your original code is fine. Thanks for the contribution!

daniel-sanche (approved):

tests/unit/handlers/test_container_engine.py
ysde: Hi @Daniel-Sanche 

Thank you!

I learned a lot from here :)

PR #209, opened 2021-03-09
fix: no duplicate logs on GCF or GAE

### LINKED ISSUE #184: tests: add duplication environment test
some platforms are sending duplicate logs. We should add a test to make sure this is caught

daniel-sanche:
Previously, the library would send duplicate logs on certain environments, due to built-in logging handlers that would write to standard output. This PR removes those extra handlers as part of the `setup_logging` function now when detected.

Also adds a test to ensure all platforms receive exactly one for each log

fixes https://github.com/googleapis/python-logging/issues/184


tests/unit/handlers/test_handlers.py
```diff
@@ -165,8 +166,60 @@ def test_setup_logging_excludes(self):
         self.assertNotIn(handler, excluded_logger.handlers)
         self.assertFalse(excluded_logger.propagate)
 
+    def test_remove_handlers_gcf(self):
```
busunkim96, line 169: I recommend using `mock.patch.dict` for temporarily modifying `os.environ`. It will handle clean up as well. https://docs.python.org/3/library/unittest.mock.html#patch-dict

Daniel-Sanche: Thanks, that looks much better

freelerobot (approved):

Issue #226, opened 2021-03-18
abstract ContainerEngineHandler into StructuredLogHandler for serverless environments


daniel-sanche:
We currently have the ContainerEngineHandler that writes formatted logs to standard output, which are then parsed and uploaded to stackdriver by fluentd

Serverless environments have issues where logs aren't always properly flushed, so they could benefit by having the same handlers.

WE should abstract ContainerEngineHandler into StructuredLogHandler, to work for other environments

daniel-sanche:
On second thought, we should just remove ContainerEngineHandler in v3.0.0. until then, we should leave it as-is for consistency

PR #228, opened 2021-03-18
feat: use standard output logs on serverless environments

daniel-sanche:
We have had bug reports that some logs aren't getting through when using the library in serverless environments, because the environment will spin down before logs are flushed.

To fix this, we can log to standard output in JSON format, and allow the environment's native log handling code to extract the logs to Stackdriver. 

This method is already used for GKE. Instead of reusing the ContainerEngineHandler, we will create a new more general StructuredLogHandler, and deprecate ContainerEngineHandler at the next major relase

Merging this branch into v2_update_2 to group work for [current milestone](https://github.com/googleapis/python-logging/milestone/4).

https://github.com/googleapis/python-logging/issues/226


google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -0,0 +1,54 @@
...
+        self.name = name
+        self.project_id = project
+
+        # add extra keys to log record
+        self.addFilter(CloudLoggingFilter(project))
+
+        # make logs appear in GCP structured logging format
+        self.formatter = logging.Formatter(GCP_FORMAT)
+
+    def format(self, record):
```
simonz130, line 45: don't you need to check that record is not null?

google/cloud/logging_v2/client.py
```diff
@@ -347,18 +350,22 @@ def get_default_handler(self, **kw):
...
-            and monitored_resource.type == _GKE_RESOURCE_TYPE
-        ):
-            return ContainerEngineHandler(**kw)
-        else:
-            return CloudLoggingHandler(self, resource=monitored_resource, **kw)
+        if isinstance(monitored_resource, Resource):
+            if monitored_resource.type == _GAE_RESOURCE_TYPE:
+                return AppEngineHandler(self, **kw)
+            elif monitored_resource.type == _GKE_RESOURCE_TYPE:
+                return ContainerEngineHandler(**kw)
```
simonz130, line 357: how is ContainerEngineHandler different from StructuredLogHandler?

google/cloud/logging_v2/client.py
```diff
@@ -347,18 +350,22 @@ def get_default_handler(self, **kw):
...
+            elif (
+                monitored_resource.type == _GCF_RESOURCE_TYPE
+                and sys.version_info[0] == 3
+                and sys.version_info[1] >= 8
+            ):
+                # Cloud Functions with runtimes > 3.8 supports structured logs on standard out
+                # 3.7 should use the standard CloudLoggingHandler
```
simonz130, line 364: probably worth mentioning what standard CloudLoggingHandler does compared to StructuredLogHandler.

google/cloud/logging_v2/handlers/__init__.py
```diff
@@ -16,12 +16,14 @@
...
+from google.cloud.logging_v2.handlers.structured_log import StructuredLogHandler
 from google.cloud.logging_v2.handlers.handlers import CloudLoggingHandler
 from google.cloud.logging_v2.handlers.handlers import setup_logging
 
 __all__ = [
     "AppEngineHandler",
     "CloudLoggingHandler",
     "ContainerEngineHandler",
+    "StructuredLogHandler",
     "setup_logging",
```
simonz130, line 28: Worth adding a comment to explain semantics of functionality that each handler adds.

google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -18,13 +18,42 @@
...
+        inferred_http, inferred_trace = get_request_data()
+        if inferred_trace is not None and self.project is not None:
+            inferred_trace = f"projects/{self.project}/traces/{inferred_trace}"
+
+        record.trace = getattr(record, "trace", inferred_trace) or ""
+        record.http_request = getattr(record, "http_request", inferred_http) or {}
+        record.request_method = record.http_request.get("requestMethod", "")
+        record.request_url = record.http_request.get("requestUrl", "")
+        record.user_agent = record.http_request.get("userAgent", "")
+        record.protocol = record.http_request.get("protocol", "")
```
simonz130, line 54: are you planning to do anything about log severity too?

daniel-sanche: This "Filter" class is to add new fields to the LogRecord object, that we can then use later when exporting logs in the Handler classes. Common fields like "severity" already exist in the LogRecord by default, so no need to add them here

google/cloud/logging_v2/handlers/structured_log.py
daniel-sanche: That shouldn't be necessary. We're just delegating the formatting work to another standard Python formatter object. If there were any error cases to worry about, the function we're calling can handle it.

google/cloud/logging_v2/handlers/__init__.py
daniel-sanche: I don't believe it's standard to use `__init__.py` for documentation.This is just an internal file used for importing packages. It's likely to get out of date if things change in the future, so I'd prefer to keep documentation alongside code

google/cloud/logging_v2/client.py
daniel-sanche: good point, added

google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -0,0 +1,55 @@
...
+
+import logging.handlers
+
+from google.cloud.logging_v2.handlers.handlers import CloudLoggingFilter
+
+GCP_FORMAT = '{"message": "%(message)s", "severity": "%(levelname)s", "logging.googleapis.com/trace": "%(trace)s", "logging.googleapis.com/sourceLocation": { "file": "%(pathname)s", "line": "%(lineno)d", "function": "%(funcName)s"}, "httpRequest": {"requestMethod": "%(request_method)s", "requestUrl": "%(request_url)s", "userAgent": "%(user_agent)s", "protocol": "%(protocol)s"} }'
+
+
+class StructuredLogHandler(logging.StreamHandler):
+    """Handler to format logs into the Cloud Logging structured log format,
```
nicoleczhu, line 26: Maybe add a link to https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity or another appropriate link, so they know "structured log format" is a formally defined hting

google/cloud/logging_v2/handlers/handlers.py
nicoleczhu: are you planning to add other httprequest fields [from here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest) as well. Nonblocking comment.

freelerobot (approved):
Worth adding samples for this. Also to more thoroughly document this in python-logging docs.

google/cloud/logging_v2/handlers/handlers.py
daniel-sanche: eventually, but some of them are more complicated, and I want to make sure they work consistently across environments. I want to put in the basics first

daniel-sanche:
> Worth adding samples for this. Also to more thoroughly document this in python-logging docs.

I have an open bug for re-doing the samples/usage guide in v3.0.0. I expect more changes before then

Issue #233, opened 2021-03-23
*Breaking* changes to the default logging resource type


jceresini:
We have been using this client to write stackdriver logs for a 2-3 years. Those logs have been written as a `global` resource type. 

We have Stackdriver logging sinks exporting those logs to a Pub/Sub topic, with sink filters for the resource.type (among other things). Another application is collecting those logs for analytics.

We redeployed the application with the latest client a few days ago, and the aforementioned Pub/Sub topic stopped receiving messages. I just tracked it down to a change in behavior in this client (we didn't notice it at first, and when we did tracking it down was difficult).

Logs from the current client are now logging stored with a resource.type of `k8s_container`.

It looks like the change is in this PR https://github.com/googleapis/python-logging/pull/200

#### Steps to reproduce

  1. Run the code below using google-cloud-logging==2.2.0 in GKE
  2. Upgrade google-cloud-logging to 2.3.0
  3. Run the code in GKE again
  4. Notice the resource.type changes from `global` to `k8s_container`

#### Code example

```python
import google.cloud.logging

client = google.cloud.logging.Client(project='test-project')
logger = client.logger('test-log')
logger.log_struct(dict(foo='bar'))
```

The resource.type is indexed in cloud logging, and used in possibly every example log filter in the google docs. I have to believe others will be affected by this breaking change.

See, for example: https://cloud.google.com/logging/docs/view/advanced-queries

Daniel-Sanche:
Thanks for reporting this, you're right, for compatibility reasons we should have kept the `resource.type` consistent until the next breaking `3.0.0` release. I'll put together a fix today to bring it back to the old behaviour for now.

That said, resource_type `k8s_container` is the expected type for code running on GKE. Is there a reason your pipeline expects a `global` type? I want to make sure I fully understand the issue

jceresini:
Simple answer: its the default behavior.

I also assumed that the `k8s_container` type was just for stdout/stderr logs from our running containers. In this case we have an application (running in GKE) that uses the logger to send structured logs to another project (that doesn't contain a GKE cluster).

The examples referenced in the [google cloud documentation](https://cloud.google.com/logging/docs/quickstart-python) don't cover using a different MonitoredResourceDescriptor. The contents of the `snippets.py` they reference (which is in this same repo) are:

```
def write_entry(logger_name):
    """Writes log entries to the given logger."""
    logging_client = logging.Client()

    # This log can be found in the Cloud Logging console under 'Custom Logs'.
[...]

PR #200, opened 2021-02-27
feat: detect monitored resources on all GCP environments

daniel-sanche:
This PR adds monitored resource detection to all current GCP platforms (GAE, GKE, GCE, GCF, Cloud Run). Previously, we only supported GAE and GKE

I accomplished this by creating a new `_monitored_resources.py` file that builds canonical resources for each environment, and contains a function to detect the active environment. It should be pretty straight forward, and won't be a disruptive change for existing users

I added new environment tests to check the resource on each environment, and unit tests to test the new file

https://github.com/googleapis/python-logging/issues/183
https://github.com/googleapis/python-logging/issues/57
https://github.com/googleapis/python-logging/issues/7
https://github.com/googleapis/python-logging/issues/10
https://github.com/googleapis/python-logging/issues/201

PR #237, opened 2021-03-23
fix: revert default resource behavior to avoid breaking changes

daniel-sanche:
We recently added some changes to better auto-detect monitored resources in the environment (https://github.com/googleapis/python-logging/pull/200, https://github.com/googleapis/python-logging/pull/207). unfortunately, some of these changes ended up breaking existing workflows, which relied on a defailt `global` resource (https://github.com/googleapis/python-logging/issues/233).

This PR reverts auto-detection in the case where the API previously provided a default behavior. An auto-detected `resource` will still be added when using `setup_logging`, as that is in line with the existing API contract.

This PR will be reverted at version v3.0.0 to make resource auto-detection the standard

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -16,6 +16,8 @@
 
 import logging
 
+
+from google.cloud.logging_v2.logger import _GLOBAL_RESOURCE
 from google.cloud.logging_v2.handlers.transports import BackgroundThreadTransport
 from google.cloud.logging_v2.handlers._monitored_resources import detect_resource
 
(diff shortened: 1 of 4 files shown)


google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -80,15 +82,12 @@ def __init__(
                 :class:`.BackgroundThreadTransport`. The other
                 option is :class:`.SyncTransport`.
             resource (~logging_v2.resource.Resource):
-                Resource for this Handler. If not given, will be inferred from the environment.
+                Resource for this Handler. Defaults to ``GLOBAL_RESOURCE``.
```
nicoleczhu, line 85: Nit: Defaults to ``_GLOBAL_RESOURCE```

freelerobot (approved):

PR #258, opened 2021-04-15
feat: Improve source location overrides

daniel-sanche:
- I improved how source_entries are handled in the CloudLoggingFilter, to better match the LogEntry spec
- I added a bunch of tests around user overrides for `http_request`, `trace`, `resource`, and `source_location` fields


tests/unit/handlers/test_structured_log.py
```diff
@@ -149,3 +149,50 @@ def test_format_with_request(self):
             result = json.loads(handler.format(record))
             for (key, value) in expected_payload.items():
                 self.assertEqual(value, result[key])
+
+    def test_format_overrides(self):
```
simonz130, line 153: add a note about what this test it doing in details.

PR #264, opened 2021-04-20
feat: allow custom labels with standard library logging

daniel-sanche:
The library now fully supports passing in extra labels when logging though the standard library

I also added a number of unit and environment tests around this new functionality

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -152,7 +152,7 @@ class CloudLoggingHandler(logging.StreamHandler):
         trace_id = getattr(record, "trace", None)
         span_id = getattr(record, "span_id", None)
         http_request = getattr(record, "http_request", None)
-        source_location = getattr(record "source_location", None)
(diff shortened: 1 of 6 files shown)


google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -61,6 +62,12 @@ def filter(self, record):
         inferred_http, inferred_trace = get_request_data()
         if inferred_trace is not None and self.project is not None:
             inferred_trace = f"projects/{self.project}/traces/{inferred_trace}"
+        # set labels
+        user_labels = getattr(record, "labels", {})
+        record.total_labels = {**self.default_labels, **user_labels}
+        record.total_labels_str = ", ".join(
+            [f'"{k}": "{v}"' for k, v in record.total_labels.items()]
```
nicoleczhu, line 69: just out of curiosity, would this work if users provided nested labels? actually, do we/do we not allow nested labels?

freelerobot (approved):

PR #267, opened 2021-04-22
feat: support span inference

daniel-sanche:
This PR adds automatic span detection, and allows users to override it by sending `spanId` in the `extras` dictionary.

The span format seems to be underdocument, so I based my span capture logic on [what the go library is doing](https://github.com/googleapis/google-cloud-go/blob/1dacadbbf57c5667e212af1cc2d135ff6084e6d3/logging/logging.go#L946).

### DIFF
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -121,7 +121,7 @@ def get_request_data_from_django():
 
 
 def _parse_trace_span(header):
-    """Given an X_CLOUD_TRACE header, extrace the trace and span ids.
+    """Given an X_CLOUD_TRACE header, extract the trace and span ids.
 
     Args:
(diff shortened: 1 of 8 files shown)


google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -112,32 +113,55 @@ def get_request_data_from_django():
...
-    if header:
-        trace_id = header.split("/", 1)[0]
+    trace_id, span_id = _parse_trace_span(header)
 
-    return http_request, trace_id
+    return http_request, trace_id, span_id
+
+
+def _parse_trace_span(header):
+    """Given an X_CLOUD_TRACE header, extrace the trace and span ids.
```
simonz130, line 124: nit: extract

simonz130 (approved):

PR #274, opened 2021-04-27
feat: support AuditLog and RequestLog protos

### LINKED ISSUE #16: Logging: Add support for deserializing protobuf payloads.
Cloud logging currently does not parse logging proto message payloads. 

This forces items like GAE RequestLogs to appear serialized under ```payload_pb``` as shown below

```
<google.cloud.logging.entries.ProtobufEntry object at 0x7fa6793621d0>

{
'payload_pb': type_url: "type.googleapis.com/google.appengine.logging.v1.RequestLog"
value: "\n\025s~mineral-minutia-820\022\0011\032T5b26e25700ff0a0425ac3c89260001737e6d696e6572616c2d6d696e757469612d383230000131000100\"\013216.9.110.82\014\010\327\304\233\331\005\020\210\341\200\271\002:\014\010\327\304\233\331\005\020\270\316\373\272\002B\005\020\260\355\372\001R\003GETZ\017/js/myjsfile.jsb\010HTTP/1.1h\260\002p_\202\001hMozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.87 Safari/537.36\242\001\037mineral-minutia-820.appspot.com\320\001\377\377\377\377\377\377\377\377\377\001\330\001\001\262\002\0061.9.54\320\002\001"
, 
'http_request': {'status': 304}, 
'resource': Resource(type=u'gae_app', labels={u'module_id': u'default', u'project_id': u'mineral-minutia-820', u'zone': u'us3', u'version_id': u'1'}), 
'severity': None, 
'insert_id': u'5b26e257000da4438cff347b', 
'timestamp': datetime.datetime(2018, 6, 17, 22, 36, 7, 656421, tzinfo=<UTC>), 
'labels': {u'clone_id': u''}, 
'logger': <google.cloud.logging.logger.Logger object at 0x7fa67b204cd0>, 
'payload': None
}

```
The ```RequestLog``` should get deserialized and placed into the ```payload``` field so its usable

- type.googleapis.com/google.cloud.audit.AuditLog
-  type.googleapis.com/google.appengine.logging.v1.RequestLog
   -   https://github.com/googleapis/googleapis/blob/master/google/appengine/logging/v1/request_log.proto

daniel-sanche:
If you run `logger.list_logs()` and encounter an unknown proto log, it will crash the library

This PR adds support for the [two officially supported](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) proto types:
- "type.googleapis.com/google.cloud.audit.AuditLog"
- "type.googleapis.com/google.appengine.logging.v1.RequestLog"

Fixes https://github.com/googleapis/python-logging/issues/16


tests/system/test_system.py
```diff
@@ -142,32 +145,121 @@ def tearDown(self):
...
+        elif isinstance(data, dict):
+            return Value(struct_value=TestLogging._dict_to_struct(data))
+        else:
+            raise TypeError("Unknown data type: %r" % type(data))
+
+    @staticmethod
+    def _dict_to_struct(data):
+        return Struct(fields={k: TestLogging._to_value(v) for k, v in data.items()})
+
+    def test_list_entry_with_auditlog(self):
```
simonz130, line 171: worth adding a comment describing the test case. Current name is too generic to explain what is specifically done in the test with audit logs

tests/system/test_system.py
```diff
@@ -142,32 +145,121 @@ def tearDown(self):
...
+            "serviceName": "test",
+            "status": {"code": 0},
+        }
+        audit_struct = self._dict_to_struct(audit_dict)
+
+        logger = Config.CLIENT.logger("audit-proto")
+        logger.log_proto(audit_struct)
+
+        # retrieve log
+        filter_ = self.TYPE_FILTER.format(type_url) + f" AND {_time_filter}"
```
simonz130, line 197: this is a side note for logging language syntax and not to this specific line:
I find it ugly when we concatenate strings to form a query. Perhaps, we should think about creation of an abstraction that will make forming these filters more intuitive when you read the code.
For example:

Instead of 
```
xxx + " AND " + time_filter
```
it could be 
```
x.And(new BinaryOperator(time1, operands.Equal, time2)
```
This could be a separate task and not part of this PR

tests/unit/test_entries.py
```diff
@@ -503,6 +503,20 @@ def test_to_api_repr_defaults(self):
         }
         self.assertEqual(entry.to_api_repr(), expected)
 
+    def test_to_api_repr_struct(self):
```
simonz130, line 506: please, no abbreviations in test or other method names

daniel-sanche: "to_api_repr" is an existing function that we can't change here without breaking the API. We could add this to the list for v3.0.0 changes (but it may not be worth the user confusion)

tests/system/test_system.py
daniel-sanche: That's an interesting idea. I'm not sure if that specific example would be seen as idiomatic for Python, but there's probably something interesting we could do with this

nicoleczhu: 2 cents: I recall in Python (and Node), this kind of interpolation is preferred

freelerobot (approved):

PR #292, opened 2021-05-11
fix: improve API compatibility for next release

daniel-sanche:
- As aprt of the next release, I removed some fields in the http_request data (referer, remoteIp, requestSize), because they don't work consistently across all GCP environments yet. I realized this would be considered a breaking change for AppEngine users who may already be relying on these fields. Instead of removing the fields, non-appengine handlers now filter them out. We can re-consider the future of these fields for v3.0.0
- StructuredLogHandler formatting breaks if users log text with quotes in it. Now, the handler will add "\" before each quote found in the message string
- The Flask request detection broke in the unit tests, so I also fixed that

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -93,7 +93,7 @@ class CloudLoggingFilter(logging.Filter):
         record._source_location_str = json.dumps(record._source_location or {})
         record._labels_str = json.dumps(record._labels or {})
         # break quotes for parsing through structured logging
-        record._msg_str = record.msg.replace('"', '\\"') or ""
+        record._msg_str = str(record.msg).replace('"', '\\"') if record.msg else ""
         return True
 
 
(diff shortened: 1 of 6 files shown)


google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -45,6 +45,10 @@ class CloudLoggingFilter(logging.Filter):
     overwritten using the `extras` argument when writing logs.
     """
 
+    # The subset of http_request fields have been tested to work consistently across GCP environments
```
nicoleczhu, line 48: Great commenting throughout this PR.

tests/unit/handlers/test_structured_log.py
```diff
@@ -104,6 +104,23 @@ def test_format_minimal(self):
...
 
+    def test_format_with_quotes(self):
+        """
+        When logging a message containing quotes, escape chars should be added
+        """
+        import logging
+        import json
+
+        handler = self._make_one()
+        message = '"test"'
```
nicoleczhu, line 115: Are other special chars escaped as well?

freelerobot (approved):

tests/unit/handlers/test_structured_log.py
daniel-sanche: I haven't seen any that need it. Let me know if you find others, I imagine it will be the same across languages

Issue #202, opened 2021-03-02
feat!: deprecate AppEngineHandler and ContainerEngineHandler


daniel-sanche:
The library originally made unique handlers for each environment. This doesn't scale well as GCP continues to add environments, especially since (other than the MonitoredResource) most of the code is shared.

As part of the breaking update, we should consider removing the legacy handlers and refactoring

Issue #245, opened 2021-03-29
Update usage guide for v3.0.0


daniel-sanche:
As part of the v3.0.0 release, we should update the usage guide. It should mention newer GCP platforms, and talk about new features

- [ ] explain deprecated classes, and give a clear path forward ([link](https://github.com/googleapis/python-logging/pull/298#discussion_r634779608))
- [ ] document `get_default_handler` ([link](https://github.com/googleapis/python-logging/pull/298#discussion_r634853652))

daniel-sanche:
Here's one comment from a customer that should be considered when updating the user guide:

```
A customer tries to correlate request and container logs as per [1] using python Cloud Logging client library.

First, the documentation [1] states:

Container logs are not automatically correlated to request logs unless you use a Cloud Logging client library.

That made him (and me) believe that Cloud Logging client library would add all needed fields to the application logs automatically, so they are correlated in the Logs Explorer. Eventually, we found out that it's not true and everything has to be done manually. Moreover, we don't have the list of needed fields for that, and samples are very simple [2] and don't cover this scenario.

By trials and errors and using some unit tests code [3] I determined that the mandatory fields are:

trace
resource.type
resource.labels And trace and resource.type fields in the child log entry (application) must be the same as in the parent log entry (http requests)
Some external resources [4] also mention the same:

The values for resource.type, resource.labels, and trace must be the same in the parent and its child(ren).

And it's not the first time when users complain about insufficient documentation [5]

So I see several issues here

[1] says that logs automatically correlated when Cloud Logging client library is used. It's not true, at least for Python library
Usage examples [2] are very basic and don't caver more advanced scenarios
It's not clear what fields have to be added/populated to correlate logs
Could we do anything regarding this? Some documentation updates/improves?

[1] https://cloud.google.com/run/docs/logging#correlate-logs
[2] https://googleapis.dev/python/logging/latest/usage.html#writing-log-entries
[3] https://github.com/googleapis/python-logging/blob/db9da3700511b5a24c3c44c9f4377705937caf46/tests/unit/test_logger.py#L305
[4] https://medium.com/google-cloud/combining-correlated-log-lines-in-google-stackdriver-dd23284aeb29
[5] https://github.com/googleapis/google-cloud-python/issues/7457
```

PR #298, opened 2021-05-13
chore!: deprecate AppEngineHandler and ContainerEngineHandler 

daniel-sanche:
This is the first PR for the upcoming 3.0.0 release. Removes AppEngineHandler and ContainerEngineHandler in favor of the more generic CloudLoggingHandler (for network logging) and StructuredLogHandler (for standard output JSON logging).

Fixes https://github.com/googleapis/python-logging/issues/202


docs/usage.rst
```diff
@@ -327,19 +327,14 @@ logging handler can use different transports. The default is
 fluentd logging handlers
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Besides :class:`~google.cloud.logging.handlers.handlers.CloudLoggingHandler`,
-which writes directly to the API, two other handlers are provided.
-:class:`~google.cloud.logging.handlers.app_engine.AppEngineHandler`, which is
-recommended when running on the Google App Engine Flexible vanilla runtimes
-(i.e. your app.yaml contains ``runtime: python``), and
```
simonz130, line 334: We should mark the classes as deprecated: update the docs to mention that they are not returned anymore and that new classes replace them. Folks who have existing code using 2.* library release should have a path forward that explains how to upgrade to 3.0.0

simonz130 (changes requested):

google/cloud/logging_v2/client.py
```diff
@@ -352,9 +350,9 @@ def get_default_handler(self, **kw):
 
         if isinstance(monitored_resource, Resource):
```
nicoleczhu, line 351: @daniel-sanche  is it possible to document the behavior of `get_default_handler` more thoroughly somewhere, including the default for each environment? I think some users will be initially confused why their logs are showing up in `stdout` and/or whichever other sinks they might be streaming their stdout logs to

docs/usage.rst
daniel-sanche: My plan was to go through and update the docs at the end of the milestone, just in case any docs changes become out of date later (https://github.com/googleapis/python-logging/issues/245)

Do you think that makes sense, or should I try to update the docs along-side changes?

google/cloud/logging_v2/client.py
daniel-sanche: I added this suggestion to the [open issue](https://github.com/googleapis/python-logging/issues/245) around updating the docs for v3.

Let me know if you have thoughts about whether docs should be updated along with each change, or as a pass at the end of the milestone

docs/usage.rst
simonz130: We can update the docs later. The point here is more about the annotation of the classes.

daniel-sanche:
closed in favor of https://github.com/googleapis/python-logging/pull/310

Issue #186, opened 2021-02-11
feat: make sure JSON structured logs are captured on all platforms


daniel-sanche:
add environment test to ensure structured logs are fully supported

daniel-sanche:
This will need to be an API breaking change, so pushing back to v3.0.0

I started a branch called support-json-logs for this. But it looks like I'll need to change the [transport contract](https://github.com/googleapis/python-logging/blob/6a1f19d9929137d4a5ec6bd4d758a16ec5b284e1/google/cloud/logging_v2/handlers/transports/background_thread.py) to accept a payload object, instead of a string message

daniel-sanche:
re-set to feature request, since it looks like the current code wasn't written to support this

Issue #263, opened 2021-04-20
feat: add base log method that infers type


daniel-sanche:
We shouldn't need custom functions for `log_text`, `log_proto`, etc. We should be able to infer it from the arguments

Issue #13, opened 2018-10-31
Logging: support sending structured logs to stackdriver via stdlib 'logging'.


yuvipanda:
google-cloud-logging version: 1.8.0

I want to send structured logs to stackdriver from Python using the standard logging module. You can integrate structured logging to the Python logging module with either https://github.com/madzak/python-json-logger or http://www.structlog.org/en/stable/. Would be great to preserve those structures as part of `jsonPayload` when sent to stackdriver. Currently, these get stringified and sent as `jsonPayload.message` instead.

Voyz:
Based on @crazystick 's answer, I use the following bodge to log custom objects as info payload. It derives more from the original `_Worker.enqueue` and supports passing custom fields.
```python
from google.cloud.logging import _helpers
from google.cloud.logging.handlers.transports.background_thread import _Worker

def my_enqueue(self, record, message, resource=None, labels=None, trace=None, span_id=None):
    queue_entry = {
        "info": {"message": message, "python_logger": record.name},
        "severity": _helpers._normalize_severity(record.levelno),
        "resource": resource,
        "labels": labels,
        "trace": trace,
        "span_id": span_id,
        "timestamp": datetime.datetime.utcfromtimestamp(record.created),
    }

    if 'custom_fields' in record:
        entry['info']['custom_fields'] = record.custom_fields

    self._queue.put_nowait(queue_entry)

_Worker.enqueue = my_enqueue
```

Then
```python
import logging
from google.cloud import logging as google_logging

logger = logging.getLogger('my_log_client)
logger.addHandler(CloudLoggingHandler(google_logging.Client(), 'my_log_client'))

logger.info('hello', extra={'custom_fields':{'foo': 1, 'bar':{'tzar':3}}})
```

Resulting in:
![image](https://user-images.githubusercontent.com/15874847/73927611-44220b00-4904-11ea-94b4-429418788880.png)

Which then makes it much easier to filter according to these custom_fields.

Naturally, it would be great if this would be naively supported - thanks in advance!

mauricepoirrier:
I tried to implement @Voyz solution on a Cloud Function and I couldn't make it work.
I think there is a strange behaviour with CF and monkey patching the module. 

```python
def my_enqueue(
    self, record, message, resource=None, labels=None, trace=None, span_id=None
):
    queue_entry = {
        "info": {
            "message": message,
            "python_logger": record.name,
            "error_fields": record.error_fields,
        },
[...]

PR #304, opened 2021-05-21
[DRAFT] Support Json logs

daniel-sanche:
This PR adds full support for JSON logs, along with standard text logs. Now, users can call `logging.error({'a':'b'})`, and they will get a JsonPayload in Cloud Logging, Or call `logging.error('test')` to receive a TextPayload

As part of this change, I added a generic `loger.log()` function, which serves as a generic entry-point instead of `logger.log_text` or `logger.log_struct`. It will infer which log function is meant based on the input type

Previously, the library would attach the python logger name as part of the JSON payload for each log. Now, that information will be attached as a label instead, giving users full control of the log payload fields

Fixes https://github.com/googleapis/python-logging/issues/186, https://github.com/googleapis/python-logging/issues/263, https://github.com/googleapis/python-logging/issues/13


daniel-sanche:
closed in favor of https://github.com/googleapis/python-logging/pull/316. Had to rebase branch

PR #315, opened 2021-06-04
feat!: Infer default resource in logger

### LINKED ISSUE #239: change default resource
Revert https://github.com/googleapis/python-logging/pull/237 at next breaking release

daniel-sanche:
We recently added functionality to the library to detect the monitored resource type based on the environment the code is running in (https://github.com/googleapis/python-logging/pull/200, https://github.com/googleapis/python-logging/pull/207). Starting with v3.0.0, this detected data should be populated by default by default when sending new logs, rather than defaulting to `global`.

This behaviour is already in place when using the logging handlers with the logging stdlib. Now calling the library methods directly will behave consistently

Fixes https://github.com/googleapis/python-logging/issues/239


google/cloud/logging_v2/logger.py
```diff
@@ -48,19 +49,23 @@ class Logger(object):
...
+            resource (Optional[~logging_v2.Resource]): a monitored resource object
+                representing the resource the code was run on. If not given, will
+                be inferred from the environment.
             labels (Optional[dict]): Mapping of default labels for entries written
                 via this logger.
 
         """
+        if not resource:
+            # infer the correct monitored resource from the local environment
+            resource = detect_resource(client.project)
```
simonz130, line 68: will this default to "global" in case the environment wasn't detected? (i.e. the code is running in dataflow job)

simonz130 (approved):
This looks like a small improvement but this is actually huge!!!

Thanks Dan!

google/cloud/logging_v2/logger.py
daniel-sanche: Yes, global is the fallback for `detect_resource`

PR #316, opened 2021-06-04
feat!: support json logs

daniel-sanche:
This PR adds full support for JSON logs, along with standard text logs. Now, users can call `logging.error({'a':'b'})`, and they will get a JsonPayload in Cloud Logging, Or call `logging.error('test')` to receive a TextPayload

As part of this change, I added a generic `loger.log()` function, which serves as a generic entry-point instead of `logger.log_text` or `logger.log_struct`. It will infer which log function is meant based on the input type

Previously, the library would attach the python logger name as part of the JSON payload for each log. Now, that information will be attached as a label instead, giving users full control of the log payload fields

Fixes https://github.com/googleapis/python-logging/issues/186, https://github.com/googleapis/python-logging/issues/263, https://github.com/googleapis/python-logging/issues/13


google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -59,12 +61,20 @@ def format(self, record):
         Returns:
             str: A JSON string formatted for GKE fluentd.
```
simonz130, line 62: nit: the comment is not up-to-date - we now do it for all environments

tests/unit/handlers/transports/test_sync.py
```diff
@@ -50,17 +50,42 @@ def test_send(self):
...
+        )
+        self.assertEqual(transport.logger.log_called_with, EXPECTED_SENT)
+
+    def test_send_struct(self):
+        from google.cloud.logging_v2.logger import _GLOBAL_RESOURCE
+        from google.cloud.logging_v2._helpers import LogSeverity
+
+        client = _Client(self.PROJECT)
+
+        stackdriver_logger_name = "python"
```
simonz130, line 70: let's not use "stackdriver" in the names of things in code.

simonz130 (approved):

google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -92,15 +93,19 @@ def filter(self, record):
         record._span_id = getattr(record, "span_id", inferred_span) or None
         record._http_request = getattr(record, "http_request", inferred_http)
         record._source_location = CloudLoggingFilter._infer_source_location(record)
-        record._labels = {**self.default_labels, **user_labels} or None
+        # add logger name as a label if possible
+        logger_label = {"python_logger": record.name} if record.name else {}
```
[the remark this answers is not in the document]
daniel-sanche: Mostly for backwards compatibility. Previously, all logs were json logs with a `message` field for user input and `python_logger` with the logger name as the other field.

I wanted to give users control over the entire payload so they could log arbitrary json objects, but I thought I should still put `python_logger` as a label in case some users were relying on it

Issue #323, opened 2021-06-14
Need to clean up Pub/Sub topic resources during tests


anguillanneuf:
Using this issue to track [how many](https://pantheon.corp.google.com/cloudpubsub/topic/list?project=python-docs-samples-tests&pageState=(%22cpsTopicList%22:(%22f%22:%22%255B%257B_22k_22_3A_22Topic%2520name_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22sink-pubsub_5C_22_22_2C_22i_22_3A_22name_22%257D%255D%22))) topics are created but  not cleaned up by the samples tests in 

https://github.com/googleapis/python-logging/blob/bce67419df6ad82d3d6d87957b47c6d4f02be1cf/samples/snippets/usage_guide.py#L253

Note that there's a hard limit of 10,000 Pub/Sub topics per project.

daniel-sanche:
Good catch, I opened https://github.com/googleapis/python-logging/pull/324 to address this. Let me know if you see any other resources that aren't being cleaned up properly

PR #324, opened 2021-06-15
test: clean up extra topics from snippet tests

daniel-sanche:
Ths PR removes topics created in the snippet test project that weren't being properly cleaned up

Fixes: https://github.com/googleapis/python-logging/issues/323

### DIFF
--- a/samples/snippets/usage_guide.py
+++ b/samples/snippets/usage_guide.py
@@ -264,13 +264,19 @@ def _sink_pubsub_setup(client):
     )  # API call
     # [END sink_topic_permissions]
 
-    return topic
+    # create callback wrapper to delete topic when done
+    class TopicDeleter():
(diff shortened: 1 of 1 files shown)


samples/snippets/usage_guide.py
```diff
@@ -264,13 +264,19 @@ def _sink_pubsub_setup(client):
     )  # API call
     # [END sink_topic_permissions]
 
-    return topic
+    # create callback wrapper to delete topic when done
+    class TopicDeleter:
+        def delete(self):
+            client.delete_topic(request={"topic": topic_path})
```
simonz130, line 270: nit: perhaps, worth adding a log line that topic X is deleted - for easier troubleshooting if tests fail in the future

simonz130 (approved):

Issue #203, opened 2021-03-02
[Logging] `page_size` appears to not work; no iterator(s)


DazWilkin:
See #136 

#### Code

`requirements.txt`:
```
google-auth==1.27.0
google-cloud-logging==2.2.0
```
And `main.py` (derived from the Google published [sample](https://cloud.google.com/logging/docs/reference/libraries#logging_list_log_entries-python))

Using `client.list_entries`:
```python
import os

from google.cloud import logging
from google.cloud.logging import DESCENDING

client = logging.Client()

PROJECT = os.getenv("PROJECT")
resource = "projects/{project}".format(project=PROJECT)
log_name = "{resource}/logs/cloudaudit.googleapis.com%2Factivity".format(
    resource=resource)
filter = "logName=\"{log_name}\"".format(log_name=log_name)

entries = client.list_entries(filter_=filter,
                              order_by=DESCENDING,
                              page_size=3,
                              resource_names=[resource])

print("Listing entries for logger {}:".format(log_name))
for entry in entries:
    timestamp = entry.timestamp.isoformat()
    print("{}".format(timestamp))
```

Using `logger.list_entries`:
```python
import os

from google.cloud import logging
from google.cloud.logging import DESCENDING

client = logging.Client()

log_name = "cloudaudit.googleapis.com%2Factivity"
logger = client.logger(log_name)

entries = logger.list_entries(
    order_by=DESCENDING,
    page_size=3,
)

print("Listing entries for logger {}:".format(log_name))
for entry in entries:
    timestamp = entry.timestamp.isoformat()
    print("{}".format(timestamp))
```

#### Steps to reproduce

Running the Python:

```bash
python3 cloud/main.py 
Listing entries for logger projects/dazwilkin-210302-66371557/logs/cloudaudit.googleapis.com%2Factivity:
2021-03-02T17:39:09.421835+00:00
2021-03-02T17:39:12.121825+00:00
2021-03-02T17:39:12.933201+00:00
2021-03-02T17:39:16.011850+00:00
2021-03-02T17:39:17.114823+00:00
2021-03-02T17:39:18.699000+00:00
2021-03-02T17:41:29.759709+00:00
2021-03-02T17:41:30.937346+00:00
2021-03-02T17:41:32.653000+00:00
```

> **NOTE** Neither `DESCENDING` nor limited

Running the equivalent (!) `gcloud` yields the expected result:

```bash
[...]

Issue #136, opened 2020-12-16
Arguments for list_entries are confusing


daniel-sanche:
[client.list_entries](https://googleapis.dev/python/logging/latest/_modules/google/cloud/logging_v2/client.html#Client.list_entries) takes in page_size and page_token arguments. It's not clear to me what a page_token should look like, and page size doesn't seem to have an affect (it says it should determine the "maximum number of entries to return", but that doesn't seem true).

We should look into these and make sure they are well tested and documented

b/174842529

bbayles:
Another problem: if an error is encountered during paging, there seems to be no way to resume from where you were. `list_entries` returns a generator now, which doesn't allow access to the `page_token`.

Daniel-Sanche:
What kind of error are you running into? Do you have any code that can replicate the issue?

bbayles:
Suppose I do:

```python
for flow_entry in logging_client.list_entries(
    filter_='whatever filters',
    resource_names=[f'projects/whatever-project'],
):
    pass
```

This handles paging for me, but if one of the page calls returns a 429 (slow down), I'm stuck - I can't resume paging from where I was without stating over from the first page.

Or can I? I would love to be wrong on this. If I am, how would I get the `page_token`?

bbayles:
To say it more generally: `list_entries` does paging for me. Suppose that one of the calls during paging fails *for any reason*. Does the current implementation of `list_entries` allow me to recover without starting over? If it does, it's not at all clear from the docs or source.

Daniel-Sanche:
That's a good question. I'm going to try to look into improving the way `list_entries` is handled when I have a chance. unfortunately since it's an API breaking change, any changes will have to wait for the next major release, and likely not this quarter. 

If this is causing issues in your work now, you can look into using the low-level auto-generated [`LoggingServiceV2Client.list_log_entries`](https://github.com/googleapis/python-logging/blob/c34a6211e1fa064e2a725007bf974a4bc7e0a153/google/cloud/logging_v2/services/logging_service_v2/client.py#L602) directly. It's not as idiomatic and harder to work with than the regular `logging_client.list_entries`, but it should give you more control

bbayles:
Many thanks. Could you explain succinctly what `list_log_entries` is vs. `list_entries`? The latter does not actually call the former, from what I can tell.

For what it's worth, the old version of this library (before v2) worked!

Daniel-Sanche:
[...]

PR #375, opened 2021-08-13
fix!: api consistency between HTTP and Gapic layers

### LINKED ISSUE #154: API differences between HTTP and gRPC implementations
We have two methods of interacting with the API backend: gRPC (default) and HTTP (used when client is created with `google.cloud.logging.Client(_use_grpc=0)` or `DISABLE_GRPC` is set).

This is implemented using two _*API implementations: One in [_http.py](https://github.com/googleapis/python-logging/blob/ff90fd2fb54c612fe6ab29708a2d5d984f60dea7/google/cloud/logging_v2/_http.py) and the other in [_gapic.py](https://github.com/googleapis/python-logging/blob/master/google/cloud/logging_v2/_gapic.py). It looks like in some cases, they return different values (_http returns a page iterator object directly, while _gapic often converts it to a generator. This confuses users who expect high-level APIs to act consistently regardless of the implementation used.

I suggest we refactor _http.py to be consistent with _gapic.py, and add tests to make sure they act consistently

daniel-sanche:
This PR is for the staged v3.0.0 branch

We have two separate transport options for this library: [gRPC](https://github.com/googleapis/python-logging/blob/28d141d0e8ed4560d2e33f8de0d43b0825a7f33f/google/cloud/logging_v2/_gapic.py) and [HTTP](https://github.com/googleapis/python-logging/blob/28d141d0e8ed4560d2e33f8de0d43b0825a7f33f/google/cloud/logging_v2/_http.py). They expose the same interface, but previously returned different types.

This PR wraps the HTTP `list_*` functions in a python generator, consistent with how the gRPC version works. Additionally, it adds a new `max_results` argument, so users can decide how many objects they want to return before stopping. This new argument will work consistently for both options.

Fixes https://github.com/googleapis/python-logging/issues/154, https://github.com/googleapis/python-logging/issues/136, https://github.com/googleapis/python-logging/issues/203

### DIFF
--- a/google/cloud/logging_v2/_gapic.py
+++ b/google/cloud/logging_v2/_gapic.py
@@ -100,7 +100,7 @@ class _LoggingAPI(object):
         loggers = {}
 
         if max_results is not None and max_results < 0:
-            raise ValueError('max_results must be positive')
+            raise ValueError("max_results must be positive")
 
         # create generator
         def log_entries_pager(log_iter):
@@ -217,7 +217,7 @@ class _SinksAPI(object):
         sink_iter = iter(response)
 
         if max_results is not None and max_results < 0:
-            raise ValueError('max_results must be positive')
+            raise ValueError("max_results must be positive")
 
         def sinks_pager(sink_iter):
             i = 0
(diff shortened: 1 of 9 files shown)


losalex (approved):
LGTM

PR #392, opened 2021-09-07
[DRAFT] feat: add cloud build resource detection

daniel-sanche:
The library will now auto-detect when it is running in a cloud build environment, and send the proper monitored resource metadata as part of the LogEntry


daniel-sanche:
Closing as we are deprioritizing this integration

PR #394, opened 2021-09-11
fix: exception log message format

### LINKED ISSUE #382: StructuredLogHandler appends Traceback
The StructuredLogHandler appends a Traceback after the json structured log message.
This causes a second unstructured log entry with just the traceback when the handler is used in a Google Cloud Function.

#### Code example

```python
import logging

from google.cloud.logging.handlers import StructuredLogHandler

logger = logging.getLogger()
logger.handlers.clear()
logger.addHandler(StructuredLogHandler())

try:
    raise Exception("Test")
except Exception:
    logger.exception("Failed")
```

#### Output
```
{"message": "Failed\nTraceback (most recent call last):\n  File \"main.py\", line 10, in <module>\n    raise Exception(\"Test\")\nException: Test", "severity": "ERROR", "logging.googleapis.com/labels": {}, "logging.googleapis.com/trace": "", "logging.googleapis.com/spanId": "", "logging.googleapis.com/sourceLocation": {"line": 12, "file": "main.py", "function": "<module>"}, "httpRequest": {} }
Traceback (most recent call last):
  File "main.py", line 10, in <module>
    raise Exception("Test")
Exception: Test
```

The traceback appears twice. Once inside the structured log message and again after it. This happens because a logging.Formatter is used which always appends the traceback if exc_info or exc_text is present.

daniel-sanche:
Previously, logging an exception with `logger.exception` would append traceback information as part of the log message, making it hard to read. This PR removes the extra exception info from the LogEntry

Fixes https://github.com/googleapis/python-logging/issues/382 🦕

### DIFF
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -15,7 +15,6 @@
 """Logging handler for printing formatted structured logs to standard output.
 """
 import json
-from copy import copy
 import logging.handlers
 
 from google.cloud.logging_v2.handlers.handlers import CloudLoggingFilter
@@ -60,9 +59,6 @@ class StructuredLogHandler(logging.StreamHandler):
         Returns:
             str: A JSON string formatted for GKE fluentd.
         """
-        # remove unwanted fields
-        record = copy(record)
-        record.exc_info = None
         # let other formatters alter the message
         super_payload = None
         if record.msg:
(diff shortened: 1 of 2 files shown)


google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -59,9 +60,13 @@ def format(self, record):
         Returns:
             str: A JSON string formatted for GKE fluentd.
         """
+        # remove unwanted fields
+        record = copy(record)
+        record.exc_info = None
```
minherz, line 65: since the record object is used for logging, what is a reason to create a full copy of the record object? why not to reset the .exc_info on the argument directly?
Otherwise, consider to use a distinct variable for the copy instead of "reusing" `record`.

minherz (approved):
From the change list it is hard to understand the justification for duplicating the record before formatting. Since the expected use of the record is for logging only, full duplication of the object seems unnecessary.

google/cloud/logging_v2/handlers/structured_log.py
daniel-sanche: I was doing some experimenting, but ended up modifying the record object directly

PR #406, opened 2021-09-27
chore: exclude 'CODEOWNERS' from templated files

tseaver:
See: https://github.com/googleapis/synthtool/pull/1201


parthea:
The CODEOWNERS file in this repo looks straight forward and the codeowner team is listed in `.repo-metadata.json`. We should be able to use the templated CODEOWNERS file for this repo.

PR #434, opened 2021-10-27
feat: use structured logging on GCF with python 3.7

daniel-sanche:
Previously, GCF Python 3.7 didn't support structured logging, so the library would default to the grpc CloudLoggingHandler. now that the environment supports structured logging, we can update to the StructuredLoggingHandler to make it in line with other serverless environments


losalex (approved):
LGTM

losalex (approved):
LGTM

Issue #399, opened 2021-09-20
It doesn't appear possible to list organization or folder audit log entries


quantumew:
It doesn't appear possible to list organization or folder audit log entries. I am hoping to use this SDK to poll all audit logs 
 for all orgs, folders, and projects, then push them to a stream for consumption by our log aggregation system. 

#### Environment details

  - OS type and version: AWS Lambda
  - Python version: 3.8.10
  - pip version: 21.2.4
  - `google-cloud-logging` version: 2.6.0

#### Steps to reproduce

  1. Retrieve logs for an organization or folder
  2. Iterate through pages, it results in a ValueError

#### Code example

```python
client = Client(project=GCP_ADMIN_PROJECT, _use_grpc=0)
iterator = client.list_entries(
       resource_names=["folders/xxx"],        
       order_by=ASCENDING,
       page_size=200,
)
```

#### Stack trace
```
{
  "errorMessage": "path \"folders/xxxx/logs/cloudaudit.googleapis.com%2Fdata_access\" did not match expected pattern \"\n    projects/            # static prefix\n    (?P<project>[^/]+)   # initial letter, wordchars + hyphen\n    /logs/               # static midfix\n    (?P<name>[^/]+)      # initial letter, wordchars + allowed punc\n\"",
  "errorType": "ValueError",
  "stackTrace": [
    "  File \"/var/task/lambda_function.py\", line 32, in lambda_handler\n    run(resource_id)\n",
    "  File \"/var/task/timeout_decorator/timeout_decorator.py\", line 82, in new_function\n    return function(*args, **kwargs)\n",
    "  File \"/var/task/lambda_function.py\", line 49, in run\n    flush_logs(resource_id, bookmarks)\n",
    "  File \"/var/task/lambda_function.py\", line 80, in flush_logs\n    for entry in list(page):\n",
    "  File \"/var/task/google/api_core/page_iterator.py\", line 131, in __next__\n    result = self._item_to_value(self._parent, item)\n",
    "  File \"/var/task/google/cloud/logging_v2/_http.py\", line 496, in _item_to_entry\n    return entry_from_resource(resource, iterator.client, loggers)\n",
    "  File \"/var/task/google/cloud/logging_v2/_helpers.py\", line 87, in entry_from_resource\n    return ProtobufEntry.from_api_repr(resource, client, loggers=loggers)\n",
[...]

PR #443, opened 2021-11-11
fix: allow non-project resources when parsing entries

daniel-sanche:
Previously, the entry-parsing code assumed all logs originated from a specific project, when they could originate from a `folder`, `organization`, or `billingAccount`. This PR changes the regex check to be more permissive, and adds extra test cases

Fixes https://github.com/googleapis/python-logging/issues/399 🦕


daniel-sanche:
Actually after giving this some more thought, I think this is the wrong approach. It looks like `Logger` objects must be associated with a specific project path, but some logs aren't. I think it would be safer to keep the `logger` field as None in these cases.

Closing this for now, but may re-open after doing some more research

PR #444, opened 2021-11-12
fix: allow reading logs from non-project paths

### LINKED ISSUE #399: It doesn't appear possible to list organization or folder audit log entries

daniel-sanche:
Previously, the entry-parsing code assumed all logs originated from a project, but the list_logs API allows reading from folders, organizations, and billingIds as well. If a user attempted to read from one of these non-project paths, the library would fail to build the LogEntry object and crash. This PR fixes that issue, and is more forgiving in its entry parsing logic.

Now, if a log is found with a parent path that does not represent the active project, the `entry.logger` field will be `None`.

Fixes #399

### DIFF
--- a/google/cloud/logging_v2/entries.py
+++ b/google/cloud/logging_v2/entries.py
@@ -46,11 +46,12 @@ _LOGGER_TEMPLATE = re.compile(
 )
 
 
-def logger_name_from_path(path):
+def logger_name_from_path(path, project=None):
     """Validate a logger URI path and get the logger name.
 
     Args:
         path (str): URI path for a logger API request
+        project (str): The project the path is expected to belong to
 
     Returns:
         str: Logger name parsed from ``path``.
@@ -59,7 +60,7 @@ def logger_name_from_path(path):
         ValueError: If the ``path`` is ill-formed of if the project
             from ``path`` does not agree with the ``project`` passed in.
     """
-    return _name_from_project_path(path, None, _LOGGER_TEMPLATE)
+    return _name_from_project_path(path, project, _LOGGER_TEMPLATE)
 
 
 def _int_or_none(value):
@@ -165,8 +166,13 @@ class LogEntry(_LogEntryTuple):
         logger_fullname = resource["logName"]
         logger = loggers.get(logger_fullname)
         if logger is None:
-            logger_name = logger_name_from_path(logger_fullname)
-            logger = loggers[logger_fullname] = client.logger(logger_name)
+            # attempt to create a logger if possible
+            try:
+                logger_name = logger_name_from_path(logger_fullname, client.project)
+                logger = loggers[logger_fullname] = client.logger(logger_name)
+            except ValueError:
+                # log name is not scoped to a project. Leave logger as None
+                pass
         payload = cls._extract_payload(resource)
         insert_id = resource.get("insertId")
         timestamp = resource.get("timestamp")
--- a/google/cloud/logging_v2/entries.py
+++ b/google/cloud/logging_v2/entries.py
@@ -156,7 +156,8 @@ class LogEntry(_LogEntryTuple):
                 Client which holds credentials and project configuration.
             loggers (Optional[dict]):
                 A mapping of logger fullnames -> loggers.  If not
-                passed, the entry will have a newly-created logger.
+                passed, the entry will have a newly-created logger if possible,
+                or an empty logger field if not.
 
         Returns:
             google.cloud.logging.entries.LogEntry: Log entry parsed from ``resource``.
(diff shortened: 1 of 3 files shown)


google/cloud/logging_v2/entries.py
```diff
@@ -46,11 +46,12 @@
 )
 
 
-def logger_name_from_path(path):
+def logger_name_from_path(path, project=None):
```
losalex, line 49: nit: Perhaps I am not well familiar with Python, but is there a way to create another `logger_name_from_path` function with `project` parameter? I think that it would be better then having parameters with default values

google/cloud/logging_v2/entries.py
```diff
@@ -165,8 +167,13 @@ def from_api_repr(cls, resource, client, *, loggers=None):
...
         logger = loggers.get(logger_fullname)
         if logger is None:
-            logger_name = logger_name_from_path(logger_fullname)
-            logger = loggers[logger_fullname] = client.logger(logger_name)
+            # attempt to create a logger if possible
+            try:
+                logger_name = logger_name_from_path(logger_fullname, client.project)
+                logger = loggers[logger_fullname] = client.logger(logger_name)
+            except ValueError:
+                # log name is not scoped to a project. Leave logger as None
```
losalex, line 175: Is this comment true? Seems that logger is initialized with value returned from `loggers.get(logger_fullname)` call...

losalex (approved):
Left some minor comments

google/cloud/logging_v2/entries.py
daniel-sanche: I think this is the Pythonic way to do it.  Especially since this function is just directly calling another function. If it wasn't a public function already in use, I would prefer to remove this one altogether

What's the argument against default parameters?

google/cloud/logging_v2/entries.py
daniel-sanche: `loggers.get(logger_fullname)` will return `None` if the loggers dict doesn't contain `logger_fullname`

If it's `None`. Lines 171 to 173 will try to contruct a new logger associated with `logger_name`, which will work if `logger_name` is a project id. If it's an organization or a folder, it will throw an exception. In that case, we can just leave the logger as empty and move on

Issue #411, opened 2021-10-04
Accept JSON fields from "extra" argument


m-gentil:
**Is your feature request related to a problem? Please describe.**

When attaching the [CloudLoggingHandler](https://googleapis.dev/python/logging/2.6.0/handlers.html#google.cloud.logging_v2.handlers.handlers.CloudLoggingHandler) to a standard python logger, it is not possible to perform structured logging through the `extra` keyword argument for each logging call. Please refer to the [standard logging python library](https://docs.python.org/3/library/logging.html#logging.Logger.debug) for more documentation about the `extra` parameter. Take for instance the following snippet of code, written in Python 3.9 and using `google-cloud-logging==2.6.0` as a dependency:
```python
import logging
import google.cloud.logging
from google.cloud.logging_v2.handlers import CloudLoggingHandler

client = google.cloud.logging.Client()
handler = CloudLoggingHandler(client)

cloud_logger = logging.getLogger('cloudLogger')
cloud_logger.setLevel(logging.INFO)
cloud_logger.addHandler(handler)

cloud_logger.error('bad news', extra={"error_code": 300})  # Note that we include an extra dictionary
```

The `extra` dictionary is completely lost and does not appear in the received log from the Cloud Logging UI (see screenshot).

 **Describe the solution you'd like**

All properties of a [LogRecord](https://docs.python.org/3/library/logging.html#logging.LogRecord) that are considered as `extra`s appear in the jsonPayload of the related entry in Cloud Logging. This can be implemented for instance in the body of the `emit` method of the CloudLoggingHandler ([reference](https://github.com/googleapis/python-logging/blob/6a837a5d1faab1f9fa8ac94e424e847821a0069f/google/cloud/logging_v2/handlers/handlers.py#L180)).
 
 **Describe alternatives you've considered**

The only alternative solution we found is to use the `log_struct` method of the `google.cloud.logging.logger.Logger()` object directly. This is not a valid solution for us as we make heavy use of the standard logging library shipped with python: using a separate logger would introduce too much overhead for our existing and future projects, while killing the purpose of the centralized logging infrastructure that the python standard logging library provides.

 **Additional context**
![Screenshot 2021-10-04 at 17 46 36](https://user-images.githubusercontent.com/15276897/135888964-1e37501b-c7e9-487c-b003-33499c21adb6.png)

daniel-sanche:
Hey, thanks for filing this

[...]

PR #447, opened 2021-11-16
feat: add json_fields extras argument for adding to jsonPayload

daniel-sanche:
v3.0.0 will support sending jsonPayload log entries by logging a dict object: `cloud_logger.error(my_dict)`. This feature is technically counter to the python logging spec though, which expects string inputs. For full compatibility, we advise users to send their dicts encoded as JSON like this: `cloud_logger.error(json.dumps(my_dict))`

These methods of sending JSON payloads [don't work for all customers though](https://github.com/googleapis/python-logging/issues/411), so this PR adds a third method: users can now send json data through the `extras` argument: `cloud_logger.error('bad news', extra={json_fields: my_dict}))`. Logging this way will result in a jsonPayload log, made up of the `json_fields` dict combined with the log message.

Fixes https://github.com/googleapis/python-logging/issues/411


google/cloud/logging_v2/client.py
```diff
@@ -376,7 +376,7 @@ def get_default_handler(self, **kw):
             if monitored_resource.type == _GAE_RESOURCE_TYPE:
                 return CloudLoggingHandler(self, resource=monitored_resource, **kw)
             elif monitored_resource.type == _GKE_RESOURCE_TYPE:
-                return ContainerEngineHandler(**kw)
+                return StructuredLogHandler(**kw, project_id=self.project)
```
minherz, line 379: nit: I think that I saw this commit in PR #450. Is it possible there had to be a rebase?

google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -221,9 +221,16 @@ def _format_and_parse_message(record, formatter_handler):
         record (logging.LogRecord): The record object representing the log
         formatter_handler (logging.Handler): The handler used to format the log
     """
+    passed_json_fields = getattr(record, "json_fields", {})
     # if message is a dictionary, return as-is
     if isinstance(record.msg, collections.abc.Mapping):
-        return record.msg
+        payload = record.msg
+        if passed_json_fields and isinstance(
+            passed_json_fields, collections.abc.Mapping
```
minherz, line 229: nit: if passed_json_fields is undefined (i.e. `if not passed_json_fields` then the line 226 probably will fail. should we check for it sooner?

tests/unit/handlers/test_structured_log.py
```diff
@@ -321,3 +321,26 @@ def test_format_overrides(self):
...
+    def test_format_with_json_fields(self):
+        """
+        User can add json_fields to the record, which should populate the payload
+        """
+        import logging
+        import json
+
+        handler = self._make_one()
+        message = "name: %s"
+        name_arg = "Daniel"
```
minherz, line 334: question: i am not familiar with the interface, is it supposed to format the message like with print() ?

minherz (approved):
lgtm. i did not do in-depth analysis of system and unit tests. maybe someone will look into it latter..

losalex (approved):
LGTM

google/cloud/logging_v2/client.py
daniel-sanche: Yeah, it seems like there was a merging issue in the v3.0.0 branch at some point which changed this line, and I had to fix it in both branches for tests to pass

google/cloud/logging_v2/handlers/handlers.py
daniel-sanche: I'm not sure I understand. Line 226 should be unrelated to `passed_json_fields`. It's checking if the `msg` field was populated by a dictionary. If so, lines 228-232 will attempt to add the additional `passed_json_fields` fields to the msg dictionary (if present)

tests/unit/handlers/test_structured_log.py
daniel-sanche: The idea is that Python's logging std lib [allows you to build formated log strings](https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles) like this: `logging.info("my name is: %s", "Daniel")`. This test just makes sure that our GCP library additions don't break that expected functionality

google/cloud/logging_v2/handlers/handlers.py
daniel-sanche: I tried to improve the comments here

Issue #321, opened 2021-06-10
align trace collection with nodejs library


daniel-sanche:
The Nodejs logging library was [updated to include better trace handling](https://github.com/googleapis/nodejs-logging/pull/1088/files). We should make sure Python alights with the same behaviours:
- handle the traceSampled field
- check for [traceparent header](https://www.w3.org/TR/trace-context/)

PR #450, opened 2021-11-23
feat: trace improvements

daniel-sanche:
This PR improves trace detection by adding the following features:
- adds support for the [open W3 `traceparent` header](https://www.w3.org/TR/trace-context/) format
- adds support for the `trace_sampled` field

These changes should bring the python library in line with how nodejs currently works.

I also added a number of unit, system, and environment tests to make sure the new features work as expected throughout the stack

Fixes https://github.com/googleapis/python-logging/issues/321

### DIFF
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -64,7 +64,7 @@ def get_request_data_from_flask():
     """Get http_request and trace data from flask request headers.
 
     Returns:
-        Tuple[Optional[dict], Optional[str], Optional[str]]:
+        Tuple[Optional[dict], Optional[str], Optional[str], bool]:
             Data related to the current http request, trace_id, span_id and trace_sampled
             for the request. All fields will be None if a django request isn't found.
     """
@@ -95,7 +95,7 @@ def get_request_data_from_django():
     """Get http_request and trace data from django request headers.
 
     Returns:
-        Tuple[Optional[dict], Optional[str], Optional[str], Optional[bool]]:
+        Tuple[Optional[dict], Optional[str], Optional[str], bool]:
             Data related to the current http request, trace_id, span_id, and trace_sampled
             for the request. All fields will be None if a django request isn't found.
     """
@@ -130,26 +130,31 @@ def _parse_trace_parent(header):
 
     Args:
         header (str): the string extracted from the traceparent header
+            example: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
     Returns:
-        Tuple[Optional[dict], Optional[str], Optional[bool]]:
+        Tuple[Optional[dict], Optional[str], bool]:
             The trace_id, span_id and trace_sampled extracted from the header
-            Each field will be None if not found.
+            Each field will be None if header can't be parsed in expected format.
     """
     trace_id = span_id = None
     trace_sampled = False
-    # see https://cloud.google.com/trace/docs/setup for X-Cloud-Trace_Context format
+    # see https://www.w3.org/TR/trace-context/ for W3C traceparent format
     if header:
         try:
-            VERSION_PART = r'(?!ff)[\\da-f]{2}';
-            TRACE_ID_PART = r'(?![0]{32})[\\da-f]{32}';
-            PARENT_ID_PART = r'(?![0]{16})[\\da-f]{16}';
-            FLAGS_PART = r'[\\da-f]{2}';
+            VERSION_PART = r'(?!ff)[a-f\d]{2}'
+            TRACE_ID_PART = r'(?![0]{32})[a-f\d]{32}'
+            PARENT_ID_PART = r'(?![0]{16})[a-f\d]{16}'
+            FLAGS_PART = r'[a-f\d]{2}'
             regex = f'^\\s?({VERSION_PART})-({TRACE_ID_PART})-({PARENT_ID_PART})-({FLAGS_PART})(-.*)?\\s?$'
             match = re.match(regex, header)
             trace_id = match.group(2)
             span_id = match.group(3)
-            trace_sampled = (int(match.group(5), 16) == 1)
-        except IndexError:
+            # trace-flag component is an 8-bit bit field. Read as an int
+            int_flag = int(match.group(4), 16)
+            # trace sampled is set if the right-most bit in flag component is set
+            trace_sampled = bool(int_flag & 1)
+        except (IndexError, AttributeError):
+            # could not parse header as expected. Return None
             pass
     return trace_id, span_id, trace_sampled
 
@@ -160,7 +165,7 @@ def _parse_xcloud_trace(header):
     Args:
         header (str): the string extracted from the X_CLOUD_TRACE header
     Returns:
-        Tuple[Optional[dict], Optional[str], Optional[bool]]:
+        Tuple[Optional[dict], Optional[str], bool]:
             The trace_id, span_id and trace_sampled extracted from the header
             Each field will be None if not found.
     """
@@ -186,7 +191,7 @@ def get_request_data():
     Returns:
         Tuple[Optional[dict], Optional[str], Optional[str], Optional[bool]]:
             Data related to the current http request, trace_id, span_id, and trace_sampled 
-            for the request. All fields will be None if a django request isn't found.
+            for the request. All fields will be None if a http request isn't found.
     """
     checkers = (
         get_request_data_from_django,
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -80,11 +80,11 @@ def get_request_data_from_flask():
     }
 
     # find trace id and span id
+    # first check for w3c traceparent header
     header = flask.request.headers.get(_FLASK_TRACEPARENT)
-    if header:
-        trace_id, span_id, trace_sampled = _parse_trace_parent(header)
-    else:
-        # w3 traceparent header not found. Check XCLOUD_TRACE_CONTEXT
+    trace_id, span_id, trace_sampled = _parse_trace_parent(header)
+    if trace_id is None:
+        # traceparent not found. look for xcloud_trace_context header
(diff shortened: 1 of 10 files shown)


google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -108,54 +113,94 @@ def get_request_data_from_django():
...
+        Tuple[Optional[dict], Optional[str], bool]:
+            The trace_id, span_id and trace_sampled extracted from the header
+            Each field will be None if header can't be parsed in expected format.
+    """
+    trace_id = span_id = None
+    trace_sampled = False
+    # see https://www.w3.org/TR/trace-context/ for W3C traceparent format
+    if header:
+        try:
+            VERSION_PART = r"(?!ff)[a-f\d]{2}"
```
minherz, line 144: nit: right now the version **must be** `00`. other versions are considered invalid.

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -108,54 +113,94 @@ def get_request_data_from_django():
...
+    trace_sampled = False
+    # see https://cloud.google.com/trace/docs/setup for X-Cloud-Trace_Context format
     if header:
         try:
-            split_header = header.split("/", 1)
-            trace_id = split_header[0]
-            header_suffix = split_header[1]
-            # the span is the set of alphanumeric characters after the /
-            span_id = re.findall(r"^\w+", header_suffix)[0]
+            regex = r"([\w-]+)?(\/?([\w-]+))?(;?o=(\d))?"
```
minherz, line 177: nit: according to [our doc](https://cloud.google.com/trace/docs/setup#force-trace) TRACE_ID is 32-character hexadec value that can be presented via acceptible UUID formats. SPAN_ID is a decimal value. adding it because tracing of the w3c was more strict. :slightly_smiling_face:

tests/system/test_system.py
```diff
@@ -534,6 +534,7 @@ def test_handlers_w_extras(self):
         extra = {
             "trace": "123",
             "span_id": "456",
+            "trace_sampled": True,
```
minherz, line 537: nit: do you want to add a regression test (sampled=false or omitted)?

tests/unit/handlers/test_handlers.py
```diff
@@ -154,6 +158,47 @@ def test_record_with_request(self):
...
+
+        filter_obj = self._make_one()
+        record = logging.LogRecord(None, logging.INFO, None, None, None, None, None,)
+        record.created = None
+
+        expected_path = "http://testserver/123"
+        expected_agent = "Mozilla/5.0"
+        expected_trace = "4bf92f3577b34da6a3ce929d0e0e4736"
+        expected_span = "00f067aa0ba902b7"
+        combined_trace = f"00-{expected_trace}-{expected_span}-03"
```
minherz, line 180: nit: you might want to set the flags field to `01` to keep aligned with [other flags](https://www.w3.org/TR/trace-context/#other-flags) of the standard.
on the same account would you like to add sampling=false test?

tests/unit/handlers/test_structured_log.py
```diff
@@ -267,6 +270,41 @@ def test_format_with_request(self):
...
+
+        handler = self._make_one()
+        logname = "loggername"
+        message = "hello world，嗨 世界"
+        record = logging.LogRecord(logname, logging.INFO, "", 0, message, None, None)
+        expected_path = "http://testserver/123"
+        expected_agent = "Mozilla/5.0"
+        expected_trace = "4bf92f3577b34da6a3ce929d0e0e4736"
+        expected_span = "00f067aa0ba902b7"
+        trace_header = f"00-{expected_trace}-{expected_span}-09"
```
minherz, line 285: nit: does flags=09 has a special meaning?

minherz (approved):
lgtm: high art of regex, bravo!

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -27,11 +27,13 @@
...
 _DJANGO_CONTENT_LENGTH = "CONTENT_LENGTH"
-_DJANGO_TRACE_HEADER = "HTTP_X_CLOUD_TRACE_CONTEXT"
+_DJANGO_XCLOUD_TRACE_HEADER = "HTTP_X_CLOUD_TRACE_CONTEXT"
+_DJANGO_TRACEPARENT = "HTTP_TRACEPARENT"
 _DJANGO_USERAGENT_HEADER = "HTTP_USER_AGENT"
 _DJANGO_REMOTE_ADDR_HEADER = "REMOTE_ADDR"
 _DJANGO_REFERER_HEADER = "HTTP_REFERER"
-_FLASK_TRACE_HEADER = "X_CLOUD_TRACE_CONTEXT"
+_FLASK_XCLOUD_TRACE_HEADER = "X_CLOUD_TRACE_CONTEXT"
+_FLASK_TRACEPARENT = "TRACEPARENT"
```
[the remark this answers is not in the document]
daniel-sanche: Yeah, for some reason django uses "HTTP_" as a prefix when reading most HTTP headers. 

At some point as we add more framework support, we may want to revisit this to add the prefix later in django code instead of in the constant, but for now I'd prefer to leave it as-is

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -62,13 +64,12 @@ def get_request_data_from_flask():
     """Get http_request and trace data from flask request headers.
 
     Returns:
-        Tuple[Optional[dict], Optional[str], Optional[str]]:
-            Data related to the current http request, trace_id, and span_id for
-            the request. All fields will be None if a django request isn't
-            found.
+        Tuple[Optional[dict], Optional[str], Optional[str], bool]:
```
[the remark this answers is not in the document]
daniel-sanche: No, the [LogEntry spec](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) says if the sampling decision is uncertain, you should default the `trace_sampled` field to False. So this function will always return True or False for that field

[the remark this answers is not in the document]
tseaver: FWIW, the code immediately below returns `None`.

google/cloud/logging_v2/handlers/_helpers.py
daniel-sanche: How are you handling other versions? Just ignoring anything other than `00`?

It looks like [nodejs is assuming forward-compatibility](https://github.com/googleapis/nodejs-logging/blob/005cced4f8c1fe5f90496537e4731c2272946cb7/src/context.ts#L228), based on [this part of the spec](https://www.w3.org/TR/trace-context/#versioning-of-traceparent) ("The current version of this specification assumes that future versions of the traceparent header will be additive to the current one."). I'm fine with locking to version 00 for now though if we decide that's safer. Let me know if you have thoughts.

google/cloud/logging_v2/handlers/_helpers.py
daniel-sanche: Yeah, I'm kind of curious on your thoughts here. I made the w3c code more strict, because the spec is pretty explicit about dropping invalid headers (which makes sense if you want to ensure all vendors can reliably talk to each other). 

I thought we might want to be a bit more permissible in the `X-Cloud-Trace-Context` parsing, and leave the correct formatting to the user. But maybe it would make sense to be strict for both?

tests/system/test_system.py
daniel-sanche: good point, I added a couple more assertions to this file

tests/unit/handlers/test_handlers.py
daniel-sanche: trace_sampled should be True if the last bit in the flags byte string is True. I wanted to test the more complex case of "0...011" instead of the simple case of `0...001`. (The 01 case and others are thoroughly tested elsewhere in test__helpers. test_handlers is slightly more high level)

tests/unit/handlers/test_structured_log.py
daniel-sanche: Like above, I wanted to test with a flags field with multiple bits set (`b1001`), to make it more interesting. Most of the trace_sampled tests are in the `test__helpers.py` file.

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -79,25 +80,29 @@ def get_request_data_from_flask():
     }
 
     # find trace id and span id
-    header = flask.request.headers.get(_FLASK_TRACE_HEADER)
-    trace_id, span_id = _parse_trace_span(header)
+    header = flask.request.headers.get(_FLASK_TRACEPARENT)
+    if header:
+        trace_id, span_id, trace_sampled = _parse_trace_parent(header)
```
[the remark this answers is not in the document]
daniel-sanche: How do you suggest we let the caller know about it? I don't think we should throw an exception. We could print custom logs when we find unparseable metadata, but that's not something we've really done so far in the logging library code base (as far as I know)

google/cloud/logging_v2/handlers/_helpers.py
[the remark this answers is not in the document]
daniel-sanche: Oh good catch! I started with it being optional before switching. I'll look through the code again

google/cloud/logging_v2/handlers/_helpers.py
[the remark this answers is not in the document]
losalex: What you could do then to make best effort - try to call `_parse_trace_parent` and if it fails, call `_parse_xcloud_trace`. Also if you have a way to document somehow that it is best effort, it would be a plus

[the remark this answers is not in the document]
daniel-sanche: Hmm ok, I changed the logic so it'll still work if both headers are set. It'll prefer `traceparent`, but if `traceparent` is invalid, it'll fall back to `xcloud_trace_context`.

I think the docs should make it clear all metadata is collected on a best-effort basis. I'll take a look to make sure when I update the usage guide for v3.0.0

losalex (approved):
LGTM

PR #484, opened 2022-02-07
fix: remove unnecessary detect_resource calls from CloudLoggingHandler

### LINKED ISSUE #475: Unable to bypass detect_resource/retrieve_metadata_server in 3.0.0
Hi, I'm having an issue with the additional `detect_resource` call that was added in 3.0.0. When I'm running on my dev machine, any traffic out to metadata.google.internal is black-holed, and hangs until urllib times out (1-2 mins). In 2.x, I was able to get around this behavior by either supplying the resource to `setup_logging` as a kwarg (which eliminated the need to call `detect_resource`), or by instantiating the `CloudLoggingHandler` manually.  

In 3.0.0 it seems it is no longer possible to bypass the call to `detect_resource`,  because the transport class is instantiating a logger without passing down anything except for name:

https://github.com/googleapis/python-logging/blob/5267152574b2ee96eb6f5c536a762f58bd2f886e/google/cloud/logging_v2/handlers/transports/background_thread.py#L280

Which will always result in a call to `detect_resource`:

https://github.com/googleapis/python-logging/blob/d86be6cf83c3f3b91c4fc0b2e0666b0ca1d7e248/google/cloud/logging_v2/logger.py#L71-L73

This is probably an edge case, but I figured I'd mention it because regardless of the hanging request, it is kind of weird that it calls `detect_resource` when I've already passed in a resource.

daniel-sanche:
Fixes https://github.com/googleapis/python-logging/issues/475

Previously, CLoudLoggingHandler would always call `detect_resource`, even when a resource is passed in. This was because the handler class wasn't sending the resource to the Transport class, so it would create a new logger that would have to re-detect the resource before usage (even though the resource field will be set by handlers in this case).

This PR fixes the issue by adding an optional `resource` argument to the transport classes, and making CloudLoggingHandler populate the field.

This PR also adds a timeout to detect_resource to prevent similar issues in the future.

### DIFF
--- a/google/cloud/logging_v2/_helpers.py
+++ b/google/cloud/logging_v2/_helpers.py
@@ -99,6 +99,7 @@ def retrieve_metadata_server(metadata_key, timeout=5):
             Key of the metadata which will form the url. You can
             also supply query parameters after the metadata key.
             e.g. "tags?alt=json"
+        timeout (number): number of seconds to wait for the HTTP request
 
     Returns:
         str: The value of the metadata key returned by the metadata server.
--- a/google/cloud/logging_v2/_helpers.py
+++ b/google/cloud/logging_v2/_helpers.py
@@ -89,7 +89,7 @@ def entry_from_resource(resource, client, loggers):
     return LogEntry.from_api_repr(resource, client, loggers=loggers)
 
 
-def retrieve_metadata_server(metadata_key):
+def retrieve_metadata_server(metadata_key, timeout=5):
     """Retrieve the metadata key in the metadata server.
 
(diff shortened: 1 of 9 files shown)


google/cloud/logging_v2/_helpers.py
```diff
@@ -89,7 +89,7 @@ def entry_from_resource(resource, client, loggers):
     return LogEntry.from_api_repr(resource, client, loggers=loggers)
 
 
-def retrieve_metadata_server(metadata_key):
+def retrieve_metadata_server(metadata_key, timeout=5):
```
simonz130, line 92: worth specifying the units for the timeout

simonz130 (approved):

PR #485, opened 2022-02-07
fix: fix system test for mtls

arithmetic1728:
mtls is only available for grpc clients, so we need to skip those system tests for http clients. This fixes: b/218322984


arithmetic1728:
@daniel-sanche I don't think the app engine test failure is related to my change since my change has no effect unless mTLS specific env var is set.

daniel-sanche (approved):

PR #504, opened 2022-03-14
Performance tests

daniel-sanche:


arbrown:
Hey there! We're working on fixing some failing environment tests in https://github.com/googleapis/python-logging/pull/503.  It's almost all working, just one left.  Hopefully we can merge that change in and get all the tests to pass here too.

PR #511, opened 2022-03-22
chore: set up CI on v1 branch

busunkim96:
Set up GitHub Actions, update release script on v1

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -54,4 +54,4 @@ jobs:
     - name: Report coverage results
       run: |
         coverage combine .coverage-results/.coverage*
-        coverage report --show-missing --fail-under=100
+        coverage report --show-missing --fail-under=98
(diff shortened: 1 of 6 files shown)


arbrown (approved):

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,57 @@
...
+  pull_request:
+    branches:
+      - v1
+name: unittest
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: ['3.6', '3.7', '3.8']
```
parthea, line 11: We could drop 3.8 since it's not in the noxfile in the v1 branch

https://github.com/googleapis/python-logging/blob/7809d806dac776c385339dc71844d45245f0ce9a/noxfile.py#L100-L102

```suggestion
        python: ['3.6', '3.7']
```

parthea (changes requested):

parthea (approved):

PR #564, opened 2022-06-01
fix(deps): require protobuf <4.0.0dev

parthea:
Towards b/234444818


arbrown:
Is this because the 4.x line is too unstable?  Do we have a plan to upgrade eventually? Or just leave it under 4 until there is a reason to upgrade?

arbrown (approved):

arbrown:
That sounds good to me.  Thanks!

PR #595, opened 2022-08-04
test: add performance tests

### LINKED ISSUE #594: feat: add performance tests
It would be useful to have a set of tests in place that measure the exectution speed/resource usage of the library, so we could catch any performance regressions before they go live

daniel-sanche:
This PR contributes a set of performance tests. The performance tests run a benchmark of some of the most common library usages, and will raise an error if performance significantly drops in a future update to the code. It also breaks down which code paths/functions are making up most of the execution time, so we can work to optimize the slow parts in the future

This PR modifies the noxfile, which may get overwritten by owlbot. I may need to modify it as a higher-level template

Fixes https://github.com/googleapis/python-logging/issues/594


losalex (approved):
Thanks Dan - this is a great change. The only thing that I would ask here is to add to PR some high level explanation or diagram how the tests are designed and which APIs is does tests.

PR #634, opened 2022-09-27
fix: structured log handler drops reserved fields in json_fields

### LINKED ISSUE #543: Severity concatenation when setting metadata severity
When specifying both severity as extra JSON payload and using the python std lib severity, results in a concatenation of values in Cloud Logging.

Screen Shot 2022-05-02 at 1 05 34 PM

#### Environment details

  - OS type and version: Functions Gen 2
  - Python version: 3.8
  - `google-cloud-logging` version: 3.0.0

#### Steps to reproduce

  1. Set `"severity":"error"` in metadata
  2. Use `logger.info` to write log
  3. See combination of both severity types instead of selecting one or another.

#### Code example

```python
import functions_framework
from google.cloud.logging import Client


@functions_framework.http
def structured_logging(request):
    logging_client = Client()
    logging_client.setup_logging()
    import logging
    msg = "Hello, world!"

   # Important snippet below
    metadata = {"component": "arbitrary-property", "severity":"error"}
    logging.info(msg, extra={"json_fields": metadata})

    return "Success: A log message was written"
```

daniel-sanche:
Fixes: https://github.com/googleapis/python-logging/issues/543

Adding reserved fields like `severity` to the `json_fields` for the StructuredLogHandler was causing issues, as the json fields would attempt to overwrite the fields in the LogEntry, rather than staying in the jsonPayload, as intended. 

This PR fixes this by maintaining a list of reserved field names (taken from the [structured logging spec](https://cloud.google.com/logging/docs/structured-logging)), and filtering them out of `json_fields` before printing the log

### DIFF
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -36,6 +36,13 @@ GCP_FORMAT = (
     "}"
 )
 
+GCP_STRUCTURED_LOGGING_FIELDS = frozenset({
+    "severity", "httpRequest", "time", "timestamp",
+    "timestampSeconds", "timestampNanos", "logging.googleapis.com/insertId",
+    "logging.googleapis.com/labels", "logging.googleapis.com/operation",
+    "logging.googleapis.com/sourceLocation", "logging.googleapis.com/spanId",
+    "logging.googleapis.com/trace", "logging.googleapis.com/trace_sampled"
+    })
 
 class StructuredLogHandler(logging.StreamHandler):
     """Handler to format logs into the Cloud Logging structured log format,
@@ -70,6 +77,10 @@ class StructuredLogHandler(logging.StreamHandler):
         message = _format_and_parse_message(record, super(StructuredLogHandler, self))
 
         if isinstance(message, collections.abc.Mapping):
+            # remove any special fields
+            for key in list(message.keys()):
+                if key in GCP_STRUCTURED_LOGGING_FIELDS:
+                    del message[key]
             # if input is a dictionary, encode it as a json string
             encoded_msg = json.dumps(message, ensure_ascii=False)
             # strip out open and close parentheses
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -36,6 +36,8 @@ GCP_FORMAT = (
     "}"
 )
 
+# reserved fields taken from Structured Logging documentation:
+# https://cloud.google.com/logging/docs/structured-logging
 GCP_STRUCTURED_LOGGING_FIELDS = frozenset(
     {
         "severity",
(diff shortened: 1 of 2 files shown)


google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -36,6 +36,24 @@
     "}"
 )
 
+GCP_STRUCTURED_LOGGING_FIELDS = frozenset(
```
losalex, line 39: nit: is that list described anywhere in documentation? I wonder if we could provide any links, so in future the list can be tracked

tests/unit/handlers/test_structured_log.py
```diff
@@ -207,6 +207,52 @@ def test_format_with_custom_formatter(self):
...
+    def test_format_with_reserved_json_field(self):
+        # drop json_field data with reserved names
+        # related issue: https://github.com/googleapis/python-logging/issues/543
+        import logging
+        import json
+
+        handler = self._make_one()
+        message = "hello world"
+        extra = "still here"
+        json_fields = {
```
losalex, line 219: Is this supposed to be case sensitive? I would add extra tests for fields with capital letters...

losalex (approved):
Left some minor comments

nickkieffer:
@daniel-sanche I _may_ have run into an issue with one of the recent logging changes in here 
I think I barked up the wrong tree on the Java changes made in the same release here (with my details)
https://github.com/googleapis/java-logging/pull/1027#issuecomment-1271982511

daniel-sanche:
Hey Nick, looking at the linked thread, it looks like you may not be using the `google-cloud-logging` Python library?

The code you linked looks like it's just using the standard Python logging module, printing to standard out:

```
import logging
log = logging.getLogger(__name__)
log.debug('debuggin')
```

If you want to use this GCP logging library, you need to run [this code as well:](https://cloud.google.com/logging/docs/setup/python#connecting_the_library_to_python_logging)
```
import google.cloud.logging
client.setup_logging()
```

----------

If you're using GKE though, you shouldn't actually need to use this library directly, since GKE should have logging agents running on the nodes to extract your stdout/stderr logs into Cloud Logging automatically. You may want to look into the GKE docs to see why your logs aren't showing up (maybe your cluster was set up with the logging agents disabled?) [This guide](https://cloud.google.com/blog/products/management-tools/using-logging-your-apps-running-kubernetes-engine) might be a good place to start reading.

One last thing to check: looking at [this comment](https://github.com/googleapis/java-logging/pull/1027#issuecomment-1272085593), it seems like you're looking for the logs using the GKE Log view. That page shows a filtered view of the logs coming in to your project, trying to find only the logs related to that GKE pod. You may also want to look at the full stream of logs in the main GCP Cloud Logging interface (https://console.cloud.google.com/logs). It's possible your logs **are** being ingested into Cloud Logging, but with the wrong format, and so they aren't showing up in the filtered GKE Log view

nickkieffer:
Disregard, looks like the ops team found an issue in our configuration. Sorry for the distraction!

PR #644, opened 2022-10-06
fix(deps): allow protobuf 3.19.5

parthea:


setup.py
```diff
@@ -35,7 +35,7 @@
     "google-cloud-core >= 2.0.0, <3.0.0dev",
     "grpc-google-iam-v1 >=0.12.4, <1.0.0dev",
     "proto-plus >= 1.22.0, <2.0.0dev",
-    "protobuf >= 3.20.2, <5.0.0dev",
+    "protobuf>=3.19.5,<5.0.0dev,!=3.20.0,!=3.20.1,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
```
daniel-sanche, line 38: Is there some context here? Why are these versions disallowed?

vchudnov-g: See go/tensorflow-python-client-libaries-dependency-issue internally

vchudnov-g (approved):

PR #649, opened 2022-10-11
feat: set partial_success to default to true for batched logs

### LINKED ISSUE #448: Add support for partialSuccess flag in entries.write request for oversized log entries
Today some requests might contain oversized log entries which cause entire write request to be dropped instead just of oversized log entries to be dropped (see [log-limits](https://cloud.google.com/logging/quotas#log-limits) for more information).
The solution for this problem is to add `partialSuccess` flag as described in [request-body](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write#request-body).
Custom logic to truncate oversized entries could be implemented, but unfortunately truncating JSON structured log entries is complex and building generic solution for reduction JSON data is out of scope.

daniel-sanche:
By default, if one log in a batch fails (ex: due to large size), the entire batch will be dropped. This PR sets the `partial_success` flag by default, so other logs will still be processed. We also expose `partial_success` as an optional argument to `batch.commit()`, so users can override the behavior if needed.

Note that most `logger.log` functions only typically write a single log entries, so this functionality is only strictly needed for batching. The only exception is when writing instrumentation logs, where a `logger.log` call may write two entries. For this reason, `partial_success` is always set to true for these calls

Fixes https://github.com/googleapis/python-logging/issues/448

### DIFF
--- a/google/cloud/logging_v2/_gapic.py
+++ b/google/cloud/logging_v2/_gapic.py
@@ -121,7 +121,7 @@ class _LoggingAPI(object):
         logger_name=None,
         resource=None,
         labels=None,
-        partial_success=False,
+        partial_success=True,
         dry_run=False,
     ):
         """Log an entry resource via a POST request
--- a/google/cloud/logging_v2/_http.py
+++ b/google/cloud/logging_v2/_http.py
@@ -144,7 +144,7 @@ class _LoggingAPI(object):
         logger_name=None,
         resource=None,
         labels=None,
-        partial_success=False,
+        partial_success=True,
         dry_run=False,
     ):
         """Log an entry resource via a POST request
--- a/google/cloud/logging_v2/logger.py
+++ b/google/cloud/logging_v2/logger.py
@@ -135,7 +135,6 @@ class Logger(object):
         kw["log_name"] = kw.pop("log_name", self.full_name)
         kw["labels"] = kw.pop("labels", self.labels)
         kw["resource"] = kw.pop("resource", self.default_resource)
-        partial_success = False
 
         severity = kw.get("severity", None)
         if isinstance(severity, str) and not severity.isupper():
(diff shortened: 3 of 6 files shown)


google/cloud/logging_v2/_gapic.py
```diff
@@ -121,7 +121,7 @@ def write_entries(
         logger_name=None,
         resource=None,
         labels=None,
-        partial_success=False,
```
losalex, line 124: Based on https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write, seems partialSuccess is optional - I guess keeping default to `false` would be more correct in this case, right?

losalex (commented):
Left a comment, lets have a discussion about it

google/cloud/logging_v2/_gapic.py
daniel-sanche: It's hard to say which is more correct. Based on issue https://github.com/googleapis/python-logging/issues/448, I thought you were leaning towards setting `partialSuccess` to `True` by default, since it would prevent losing entire batches of logs when one is over-size.

But there's also an argument that we should just stick to the backend defaults whenever possible (and I assume this defaults to `False`? But it's not completely clear).

I could go either way, but we should probably at least be consistent about it across languages

losalex: > It's hard to say which is more correct. Based on issue #448, I thought you were leaning towards setting `partialSuccess` to `True` by default, since it would prevent losing entire batches of logs when one is over-size.
> 
> But there's also an argument that we should just stick to the backend defaults whenever possible (and I assume this defaults to `False`? But it's not completely clear).
> 
> I could go either way, but we should probably aim to be consistent about it across languages

Oh, perhaps I was misleading - I was meaning to expose an ability for setting `partialSuccess` in configuration as we have it in other languages.

losalex (approved):

PR #653, opened 2022-10-19
feat: Generate a copy of json_fields before adding message

### LINKED ISSUE #652: Generate a copy of json_fields before adding message
### Description
We had a problem where we wanted to log a payload to another service via "json_fields", however, the logging handler added the key "message" to the payload so we started having errors. I propose to generate a copy of the json_fields instead of using the user supplied one.

### Proposed solution
Make a copy of the "json_fields" param inside extra before processing it.
payload = {"id": 1}
logging.info("hello", extra={"json_fields": payload})
# payload becomes {"id": 1, "message": "hello"}

mat105:
We had a problem where we wanted to log a payload to another service via "json_fields", however, the logging handler added the key "message" to the payload so we started having errors. I propose to generate a copy of the json_fields instead of using the user supplied one.

I made a copy before the message is added to the dictionary.

Fixes #652

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -252,6 +252,7 @@ def _format_and_parse_message(record, formatter_handler):
         pass
     # if json_fields was set, create a dictionary using that
     if passed_json_fields and isinstance(passed_json_fields, collections.abc.Mapping):
+        passed_json_fields = passed_json_fields.copy()
         if message != "None":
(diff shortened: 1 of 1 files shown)


daniel-sanche:
LGTM, although it could also use a unit test to be safe going forward.

If you sign the CLA I can merge this, but feel free to close and I can make the change as well. Thanks for putting this together!

mat105:
OK! i'll close this :)

PR #654, opened 2022-10-24
fix: json fields dictionary has modification side effect

### LINKED ISSUE #652: Generate a copy of json_fields before adding message

daniel-sanche:
Fixes https://github.com/googleapis/python-logging/issues/652

Based on this [original PR](https://github.com/googleapis/python-logging/pull/653), with some unit tests added (thanks @mat105!)

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -252,6 +252,7 @@ def _format_and_parse_message(record, formatter_handler):
         pass
     # if json_fields was set, create a dictionary using that
     if passed_json_fields and isinstance(passed_json_fields, collections.abc.Mapping):
+        passed_json_fields = passed_json_fields.copy()
         if message != "None":
             passed_json_fields["message"] = message
         return passed_json_fields
--- a/tests/unit/handlers/test_handlers.py
+++ b/tests/unit/handlers/test_handlers.py
@@ -864,6 +864,7 @@ class TestFormatAndParseMessage(unittest.TestCase):
                 value, json_fields[key], f"expected_payload[{key}] != result[{key}]"
             )
 
+
 class TestSetupLogging(unittest.TestCase):
     def _call_fut(self, handler, excludes=None):
         from google.cloud.logging.handlers import setup_logging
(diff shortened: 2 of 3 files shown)


google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -252,6 +252,7 @@ def _format_and_parse_message(record, formatter_handler):
         pass
     # if json_fields was set, create a dictionary using that
     if passed_json_fields and isinstance(passed_json_fields, collections.abc.Mapping):
+        passed_json_fields = passed_json_fields.copy()
```
losalex, line 255: Just wonder if it could have a perf impact... If yes - based on a fact that this functionality might be required only for some users, I would add just an extra function `isinstanceWithCopy` or something alike or alternatively have extra boolean parameter to function `isinstance` (with default `false`) to indicate that copy should be returned. This way you could isolate potential impact

daniel-sanche: Performance shouldn't be a factor. `dict.copy()` creates a shallow copy, so it should just be copying over a couple pointers. Some performance analysis can be found [here](https://bugs.python.org/issue31179). It seems that even copying very large payloads with 100+ fields should take <1 microsecond, which shouldn't be a bottleneck. And we'd expect most logs to have <10 fields

I don't think exposing an argument would make much sense here either. Note that the copy isn't returned to the user, it's just used internally in the library. We only make a copy to avoid side-effects

losalex: Exposing an extra parameter would make lot of sense if copy **would have a perf impact** as I mentioned in my original comment. But seems perf is not an issue here, so we good here.

losalex (approved):
LGTM

PR #657, opened 2022-10-25
feat: Add support for custom JSON encoders

### LINKED ISSUE #656: Add support to custom `JSONEncoder`
My company is moving out from [structlog](https://www.structlog.org/en/stable/) to start using this package, as it is intended to have better integration with google products. We're using `StructuredLogHandler` to send the log records to google cloud logging. But something structlog used to do out-of-the-box was the ability to automatically convert the extra data we provide to the logger to be sent to the stream, and therefore, we need to check all the calls we're making to `logger.info()`, `debug()`, `warning()`, etc. to check if the data type is supported by JSONEncoder, otherwise the call to `hadler.format()` will raise `TypeError`, stating that `Object of type PosixPath is not JSON serializable` (this is one example of error we're getting, because the default encoder cannot encode `pathlib.Path` instances) and the log record will not be pushed to gcloud logging
I'd like to be able to specify a custom class to encode the message to JSON, so the formatting is handled by the encoder, not the caller of `logger.info()` and its siblings
An alternative would be having a filter that prepares the data to be used by the handler, I'm even wrote a filter to handle that:
```python
class JsonableConverterFilter(logging.Filter):
    """Prepare log record to be JSON encoded

    :class:`google.cloud.logging_v2.handlers.StructuredLogHandler` encodes directly to default JSON encoder without
    the possibility to set up a custom encoder, so we need to hydrate the log record to not break when emitting
    (because broken logs break silently)

    The ideal solution would be to have a LogFormatter, but as GCloud's log handler formats to JSON inside the handler,
    we need to circumvent this issue with a :class:`logging.Filter`
    """

    def filter(self, record: logging.LogRecord) -> bool:
        fields: Dict[str, Any] = getattr(record, "json_fields", {})

        for key, value in fields.items():
            fields[key] = JsonableConverterFilter.prepare_field(value)

        return True

    @classmethod
    def prepare_field(cls, value: Any) -> TypeJSON:
        """Converts the provided type into a plain JSON type"""
        # basic types
        if isinstance(value, (int, str, float, bool)) or value is None:
            return value

        try:
            iterator = iter(value)
        except TypeError:
            # probably a custom class
            if not hasattr(value, "__dict__"):
                return str(value)

            iterator = iter(value.__dict__)

        # hashmaps
        if hasattr(value, "keys") and hasattr(value, "values"):
            return {key: cls.prepare_field(val) for key, val in value.items()}

        # sequences
        return [cls.prepare_field(item) for item in iterator]

```
but it seems to me I'm deviating the purpose of log filters by doing that
 **Additional context**
n/a for now

thiromi:
Add the ability to provide custom `JSONEncoder` classes to `StructuredLogHandler`, so the log caller doesn't have to convert to primitive types every time a log record is created

Fixes googleapis/python-logging#656 🦕

### DIFF
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -62,12 +62,15 @@ class StructuredLogHandler(logging.StreamHandler):
     and write them to standard output
     """
 
-    def __init__(self, *, labels=None, stream=None, project_id=None):
+    def __init__(
+        self, *, labels=None, stream=None, project_id=None, json_encoder_cls=None
+    ):
         """
         Args:
             labels (Optional[dict]): Additional labels to attach to logs.
             stream (Optional[IO]): Stream to be used by the handler.
             project (Optional[str]): Project Id associated with the logs.
+            json_encoder_cls (Optional[Type[JSONEncoder]]): Custom JSON encoder. Defaults to json.JSONEncoder
         """
         super(StructuredLogHandler, self).__init__(stream=stream)
         self.project_id = project_id
@@ -78,6 +81,8 @@ class StructuredLogHandler(logging.StreamHandler):
 
         # make logs appear in GCP structured logging format
         self._gcp_formatter = logging.Formatter(GCP_FORMAT)
+
+        self._json_encoder_cls = json_encoder_cls or json.JSONEncoder
 
     def format(self, record):
         """Format the message into structured log JSON.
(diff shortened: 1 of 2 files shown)


daniel-sanche:
Thanks for putting this together. My only hesitation is that the `CloudLoggingHandler` also doesn't support logging unserializable dictionary types, so it might make more sense to implement this in a more universal place. It would be ideal if the handlers can be as similar as possible.

That said, I'm not seeing a clear solution for `CloudLoggingHandler`, since it [uses a different method of parsing JSON](https://github.com/googleapis/python-logging/blob/061a3813e6053fbaef8b50f871864587a10bc966/google/cloud/logging_v2/_gapic.py#L561), and never serializes into strings like the `StructuredLogHandler`. It's not as simple as passing in a single argument like this.

I opened a [feature request](https://github.com/googleapis/python-logging/issues/658) to look into this in the future, but I don't think it needs to block merging this

daniel-sanche (approved):

PR #703, opened 2023-01-12
fix: instrumentation entries should not contain user labels

daniel-sanche:
Previously, instrumentation logs would inherit labels and other log metadata from the parent log. This was causing issues when the user expected logs with certain labels to conform to a set structure, which the instrumentation log would break

Instead, instrumentation logs should be static, containing only a severity field

### DIFF
--- a/google/cloud/logging_v2/_instrumentation.py
+++ b/google/cloud/logging_v2/_instrumentation.py
@@ -67,11 +67,9 @@ def _create_diagnostic_entry(name=_PYTHON_LIBRARY_NAME, version=_LIBRARY_VERSION
             _INSTRUMENTATION_SOURCE_KEY: [_get_instrumentation_source(name, version)]
         }
     }
-    active_kws = {"severity": "INFO"}
     # only keep the log_name and resource from the parent log
-    for key in ["log_name", "resource"]:
-        if key in kw.keys():
-            active_kws[key] = kw[key]
+    allow_list = ("log_name", "resource")
+    active_kws = {k: v for k, v in kw.items() if k in allow_list}
     entry = StructEntry(payload=payload, **active_kws)
     return entry
 
--- a/google/cloud/logging_v2/_instrumentation.py
+++ b/google/cloud/logging_v2/_instrumentation.py
@@ -67,8 +67,12 @@ def _create_diagnostic_entry(name=_PYTHON_LIBRARY_NAME, version=_LIBRARY_VERSION
             _INSTRUMENTATION_SOURCE_KEY: [_get_instrumentation_source(name, version)]
         }
     }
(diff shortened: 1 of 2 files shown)


google/cloud/logging_v2/_instrumentation.py
```diff
@@ -67,8 +67,12 @@ def _create_diagnostic_entry(name=_PYTHON_LIBRARY_NAME, version=_LIBRARY_VERSION
             _INSTRUMENTATION_SOURCE_KEY: [_get_instrumentation_source(name, version)]
         }
     }
-    kw["severity"] = "INFO"
-    entry = StructEntry(payload=payload, **kw)
+    active_kws = {"severity": "INFO"}
+    # only keep the log_name and resource from the parent log
+    for key in ["log_name", "resource"]:
```
daniel-sanche, line 72: @losalex Are there any other fields that should be included? Do the other languages keep the resource field?

google/cloud/logging_v2/_instrumentation.py
```diff
@@ -67,8 +67,12 @@ def _create_diagnostic_entry(name=_PYTHON_LIBRARY_NAME, version=_LIBRARY_VERSION
             _INSTRUMENTATION_SOURCE_KEY: [_get_instrumentation_source(name, version)]
         }
     }
-    kw["severity"] = "INFO"
-    entry = StructEntry(payload=payload, **kw)
+    active_kws = {"severity": "INFO"}
```
losalex, line 70: I believe the severity should be avoided, thus will remain default - that is what we have in other languages

google/cloud/logging_v2/_instrumentation.py
losalex: Those are mandatory ones and those are the only ones which should be included.

tests/unit/test__instrumentation.py
```diff
@@ -63,3 +63,16 @@ def test_truncate_long_values(self):
 
         self.assertEqual(expected_name, self._get_diagonstic_value(entry, "name"))
         self.assertEqual(expected_version, self._get_diagonstic_value(entry, "version"))
+
+    def test_drop_labels(self):
```
losalex, line 67: Would it be possible to add a test which checks that only required fields are present?

losalex (approved):
LGTM. Please see my comments regarding severity and extra test checking only required fields are present

tests/unit/test__instrumentation.py
daniel-sanche: yup, added an assert

google/cloud/logging_v2/_instrumentation.py
daniel-sanche: done

PR #705, opened 2023-01-14
fix: send StructuredLogHandler instrumentation log using an explicit logger

### LINKED ISSUE #704: bug: StructuredLogHandler instrumentation is not reliable
The StructuredLogHandler currently sends an instrumentation log like this:

`logging.info(diagnostic_object.payload)`

This uses the root logging instance, which may not contain a GCP Logging Handler, so the log may not make its way to GCP. We should make sure the instrumentation log is written in the same method as other logs

daniel-sanche:
Fixes https://github.com/googleapis/python-logging/issues/704

> The StructuredLogHandler sends an instrumentation log in the following format:
> 
> `logging.info(diagnostic_object.payload)`
> 
> This uses the root logging instance, which may not contain a GCP Logging Handler, so the log may not make its way to GCP.

This PR addresses this by obtaining a fresh logger instance, connecting the active StructuredLogHandler object as a handler, and emitting the instrumentation that way. using this method, the instrumentation log will be reliably printed to stdout in the same format as the other logs

### DIFF
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -131,4 +131,8 @@ class StructuredLogHandler(logging.StreamHandler):
     def emit_instrumentation_info(self):
         google.cloud.logging_v2._instrumentation_emitted = True
         diagnostic_object = _create_diagnostic_entry()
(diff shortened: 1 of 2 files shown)


losalex (approved):
Good catch - thanks for fixing this!

PR #742, opened 2023-04-07
feat: add sample to explicitly specifly log level in the snippet

### LINKED ISSUE #741: Add log_struct sample to set severity

Current snippets available on [the official doc](https://cloud.google.com/logging/docs/reference/libraries#logging_list_log_entries-python) only mentions very primitive use case. However, specifying log level is very common use case and that use case has tricky part (c.f. #739) to handle, so the snippets should touch it.

Change `log_entry` snippet to assign log level explicitly.

Add comments in the document on how to achieve it.

 **Additional context**

Users need to import extra module (`google.logging.type.log_severity_pb2`) just to specify log level, so the process sounds more complex that the users usually think. (eg. just passing "info" string)

ymotongpoo:

Fixes #741

### DIFF
--- a/samples/snippets/snippets.py
+++ b/samples/snippets/snippets.py
@@ -29,6 +29,8 @@ from google.cloud import logging
 # [START logging_write_log_entry]
 def write_entry(logger_name):
     """Writes log entries to the given logger."""
+    from google.logging.type import log_severity_pb2
+
     logging_client = logging.Client()
 
     # This log can be found in the Cloud Logging console under 'Custom Logs'.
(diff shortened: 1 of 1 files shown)


daniel-sanche:
I think using protos for things like this is a bad UX. Since this is a hand-written library, we should support and encourage users to use idiomatic python types instead.

`log_struct` should support string severity, like the `log_text` sample above it. If it doesn't, we should try to fix that

ymotongpoo:
Thank you @daniel-sanche for the comment.

> I think using protos for things like this is a bad UX. Since this is a hand-written library, we should support and encourage users to use idiomatic python types instead.

I agree, the fix is just to provide how to make it work without changing the library code. If you are fine, I'm happy to contribute the code to detect the log levels in `str` and covert them to proto type internally. WDYT?

daniel-sanche:
Yeah, if you're able to contribute the code fix, that would be great!

You shouldn't need to do too much manual parsing for the `str` values. The fix may be as simple as [adding this check for ensuring uppercase, like in the non-batched log](https://github.com/googleapis/python-logging/blob/440d3f634b69991100ead96d461d3ce83cb7f009/google/cloud/logging_v2/logger.py#L144). That's the main difference that stands out to me, but I haven't looked too closely at it yet

ymotongpoo:
Thanks Dan. Then, I create another pull request to add the check and convert the type.

PR #804, opened 2023-10-26
fix: Fixed object paths in autogenerated code in owlbot.py

### LINKED ISSUE #582: Fix Autogenerated Samples
The log tailing samples ([synchronous](https://github.com/googleapis/python-logging/blob/main/samples/generated_samples/logging_v2_generated_logging_service_v2_tail_log_entries_sync.py), [asynchronous](https://github.com/googleapis/python-logging/blob/main/samples/generated_samples/logging_v2_generated_logging_service_v2_tail_log_entries_async.py)) have an issue where the object paths are out of date (`logging_v2.LoggingServiceV2Client()` should be `logging_v2.services.logging_service_v2.LoggingServiceV2Client()`, etc)

We should check the autogenerated samples to make sure they are all tested

gkevinzheng:
Fix object paths by doing a find/replace operation in `owlbot.py`, since the gapic objects are not included in the `__init__.py` of `google.cloud.logging_v2`.

Fixes #582


owlbot.py
```diff
@@ -139,6 +140,110 @@ def place_before(path, text, *before_text, escape=None):
 
 python.py_samples()
 
+# For autogenerated sample code, resolve object paths by finding the specific subpackage
+# the object belongs to. This is because we leave out all autogenerated packages from the
+# __init__.py of logging_v2. For now, this is manually copy-pasted from the __all__s of each
+# subpackage's __init__.py.
```
daniel-sanche, line 146: we discussed the possibility of eventually restructuring the gapic/veneer files into different folders, which should remove the need for this fix, right?

It might be a good idea to open a bug for that and add a link to the comment here

samples/generated_samples/logging_v2_generated_config_service_v2_copy_log_entries_async.py
```diff
@@ -36,10 +36,10 @@
 
 async def sample_copy_log_entries():
     # Create a client
-    client = logging_v2.ConfigServiceV2AsyncClient()
+    client = logging_v2.services.config_service_v2.ConfigServiceV2AsyncClient()
 
     # Initialize request argument(s)
-    request = logging_v2.CopyLogEntriesRequest(
+    request = logging_v2.types.CopyLogEntriesRequest(
```
daniel-sanche, line 42: One thought: these generated samples aren't covered by tests. Have you considered adding tests for these to catch this kind of issue in the future, like the [other snippets](https://github.com/googleapis/python-logging/tree/c201482c913b5f2eac82359787511a5a92fc3406/samples/snippets)? Smoke tests could be helpful here (i.e. just executing each function and making sure nothing crashes)

There may be a simple way to programmatically load and execute each sample, but if it's bit more complicated we could open a bug for later

daniel-sanche (approved):
LGTM, with suggestions

owlbot.py
gkevinzheng: Depending on how the gapic/veneer files are structured we might need to have a different fix for replacing `logging_v2` with something like `logging_v2.gapic`, but yeah we wouldn't need a dictionary structure like it is now.

PR #817, opened 2023-11-29
test: use separate projects for samples testing

parthea:
This change is needed to resolve `ResourceExhausted` errors which appear in the build logs for samples tests.

### DIFF
--- a/samples/snippets/export_test.py
+++ b/samples/snippets/export_test.py
@@ -23,7 +23,7 @@ import pytest
 import export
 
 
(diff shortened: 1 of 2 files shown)


parthea:
Handing this over to @gkevinzheng to investigate solutions for the error
```
Cannot get legacy ACL for a bucket that has uniform bucket-level access. Read more at https://cloud.google.com/storage/docs/uniform-bucket-level-access
```

https://source.cloud.google.com/results/invocations/8438d30f-da9d-4481-ac93-47e1f2d05f92/log

PR #845, opened 2024-01-29
fix: Added placeholder kwargs to StructuredLogHandler

### LINKED ISSUE #842: Support log name in StructuredLoggingHandler

When configuring locally, i can set the logName by using `client.setup_logging(name='my-api')` and it works correctly. My logs show up as 'my-api' in the "select log name" dropdown in the log explorer. However, the moment I deploy it to GKE, it crashes stating the `name` argument is unexpected. Digging through the code, I noticed it is because, when deployed locally, it selects CloudLoggingHandler which accepts the `name` argument, but when it's deployed to GKE, it selects StructuredLoggingHandler which does *not*.  This is frustrating for a couple reasons:

1. It makes the code less portable because in order to avoid the error, the startup procedure needs to know where it's being deployed so it knows whether or not it should add the argument.
2. If i don't set it, all of the logs show up in a log named 'python' which is confusing when you have multiple python components hosted in the same GCP project all logging to 'python'.
3. I suppose I could just add a label to all logs with the application's name using `logging.getLogger('my-api')` but then that forces me to add extras to every log statement if I want to add labels to help further scope my logs.

Support the log name in StructuredLoggingHandler. For comparison, our .NET solutions don't have this problem. It behaves the same way regardless of where it's deployed.

Point 3 of the problem statement. This is additionally problematic in a large enterprise where we're trying to build tools around log routing due to the inconsistency of how logs are partitioned. It's an inconsistent experience if nothing else.

 **Additional context**

gkevinzheng:
Added `name` and `resource` as kwargs to `StructuredLogHandler` so that calls to `client.setup_logging` with these kwargs don't crash on unexpected keyword argument exceptions.

Fixes #842

### DIFF
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -20,7 +20,6 @@ import logging
 import logging.handlers
 
 from google.cloud.logging_v2.handlers.handlers import CloudLoggingFilter
-from google.cloud.logging_v2.handlers.handlers import DEFAULT_LOGGER_NAME
 from google.cloud.logging_v2.handlers.handlers import _format_and_parse_message
 import google.cloud.logging_v2
 from google.cloud.logging_v2._instrumentation import _create_diagnostic_entry
--- a/google/cloud/logging_v2/handlers/structured_log.py
+++ b/google/cloud/logging_v2/handlers/structured_log.py
@@ -20,6 +20,7 @@ import logging
 import logging.handlers
 
 from google.cloud.logging_v2.handlers.handlers import CloudLoggingFilter
+from google.cloud.logging_v2.handlers.handlers import DEFAULT_LOGGER_NAME
 from google.cloud.logging_v2.handlers.handlers import _format_and_parse_message
 import google.cloud.logging_v2
 from google.cloud.logging_v2._instrumentation import _create_diagnostic_entry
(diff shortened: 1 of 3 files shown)


google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -63,10 +64,19 @@ class StructuredLogHandler(logging.StreamHandler):
...
+        resource=None,
+        labels=None,
+        stream=None,
+        project_id=None,
+        json_encoder_cls=None
     ):
         """
         Args:
+            name (Optional[str]): Placeholder for setup_logging consistency. Does nothing.
+            resource (Optional[dict]): Placeholder for setup_logging consistency. Does nothing.
```
daniel-sanche, line 79: Why not use `kwargs` here for the unneeded arguments?

google/cloud/logging_v2/handlers/structured_log.py
```diff
@@ -63,10 +63,18 @@ class StructuredLogHandler(logging.StreamHandler):
...
+        labels=None,
+        stream=None,
+        project_id=None,
+        json_encoder_cls=None,
+        **kwargs
     ):
         """
         Args:
+            name (Optional[str]): Placeholder for setup_logging consistency. Does nothing.
+            resource (Optional[dict]): Placeholder for setup_logging consistency. Does nothing.
```
daniel-sanche, line 77: nit: you can probably remove these comments about the args. 

Maybe add a line for `**kwargs` to the Args list here if you want to mention that others are ignored, but that's optional

daniel-sanche (approved):

PR #874, opened 2024-03-26
fix: Remove extra handlers for Cloud Run on setup

pschonev:
setup_logging() does not remove the default handler for Cloud Run environments which leads to duplicate messages and traceback broken up into multiple lines in Cloud Logging

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

### DIFF
--- a/google/cloud/logging_v2/handlers/handlers.py
+++ b/google/cloud/logging_v2/handlers/handlers.py
@@ -38,7 +38,7 @@ _INTERNAL_LOGGERS = (
 )
 
 """These environments require us to remove extra handlers on setup"""
-_CLEAR_HANDLER_RESOURCE_TYPES = ("gae_app", "cloud_function")
+_CLEAR_HANDLER_RESOURCE_TYPES = ("gae_app", "cloud_function", "cloud_run_revision")
 
 """Extra trace label to be added on App Engine environments"""
 _GAE_TRACE_ID_LABEL = "appengine.googleapis.com/trace_id"


gkevinzheng:
@pschonev Could I see a small sample of how you're doing logging using the client libraries first? I'd like to confirm/reproduce the issue, as a change like this could very easily impact other customers.

pschonev:
@gkevinzheng Sure! As per the documentation I am using [automatic configuration](https://cloud.google.com/python/docs/reference/logging/latest/std-lib-integration#automatic-configuration):

```python
client = google.cloud.logging.Client(project=PROJECT_ID)
client.setup_logging()
```

Then if I do:

```python
logging.warning("This is a warning message")
```

I am getting duplicate logs in Cloud Logging, when the Cloud Run service is deployed. I also get split up traceback when logging errors:
image
image

This happens because Client.setup_logging() will [correctly add a StructuredLogHandler](https://github.com/googleapis/python-logging/blob/f3ce8760102efab3fd8004e965bd60dca7d2e111/google/cloud/logging_v2/client.py#L368-L413) for the deployed Cloud Run service, but [setup_logging()](https://github.com/googleapis/python-logging/blob/f3ce8760102efab3fd8004e965bd60dca7d2e111/google/cloud/logging_v2/handlers/handlers.py#L268-L311) never removes the default handler from the root logger for Cloud Run environments.

Because this happens in setup_logging(), [manual handler configuration](https://cloud.google.com/python/docs/reference/logging/latest/std-lib-integration#manual-handler-configuration) also does not work as expected for Cloud Run.

My current workaround is doing what Client.setup_logging() is doing manually for Cloud Run:
```python
client = google.cloud.logging.Client(project=PROJECT_ID)
logger = logging.getLogger()

if detect_resource(client.project).type == "cloud_run_revision":
    logger.handlers.clear()
    logger.addHandler(StructuredLogHandler())
```

gkevinzheng:
@pschonev Which web framework are you running this on, if any? Are you using Flask?

pschonev:
@gkevinzheng 
```toml
python = "^3.11"
gunicorn = "^21.2.0"
Flask = "^3.0.2"
google-cloud-logging = "^3.10.0"
```

I think it doesn't matter though. All I am doing manually to fix it is removing the default handler from the root logger, add the StructuredLogHandler and then I don't have the duplicated logs and the split up traceback logs anymore. And that's what setup_logging() is supposed to do.

At least that's how I understand it from the documentation.

gkevinzheng:
@pschonev In Cloud Run, I was not able to replicate this with the following code:

```
import logging
import os

from flask import Flask

import google.cloud.logging

app = Flask(__name__)
client.setup_logging()
logging.warning("This is a warning message")

@app.route("/")
def hello_world():
    """Example Hello World route."""
    name = os.environ.get("NAME", "World")
    logging.info('This is an info message')
    return f"Hello {name}!"


if __name__ == "__main__":
    app.run(debug=True, host="0.0.0.0", port=int(os.environ.get("PORT", 8080)))
```

**However**, I was able to replicate the issue if I added a `logging.info` statement before initializing the Cloud Logging client. This is because with `logging.info`, if there are no root handlers available, the `logging` module automatically initializes a `StreamHandler` object and adds it to the root handler. Something like this successfully replicates your issue:

```
import logging
import os

from flask import Flask

import google.cloud.logging

app = Flask(__name__)

logging.info('Starting Flask')
client.setup_logging()

@app.route("/")
def hello_world():
    """Example Hello World route."""
    name = os.environ.get("NAME", "World")
    logging.info('This is an info message')
    return f"Hello {name}!"


if __name__ == "__main__":
    app.run(debug=True, host="0.0.0.0", port=int(os.environ.get("PORT", 8080)))
```

I'll talk to my team about finding a way to mitigate this sort of issue in the future, if it's the problem you're experiencing, because it's pretty subtle.

pschonev:
@gkevinzheng Sorry for not providing the complete code right away. I can see now what went wrong. I called `logging.basicConfig()` before `setup_logging()` and that's why the default handler got attached. 

Everything is working properly if I don't do that, thanks! But as you said, this is quite subtle.

PR #889, opened 2024-04-30
feat: OpenTelemetry trace/spanID integration for Python handlers

gkevinzheng:
Changes made:
- Added `opentelemetry-api` as a dependency for the library.
- Changed trace/http gathering function from `get_request_data` to `get_request_and_trace_data`
- `get_request_and_trace_data` extracts and returns OpenTelemetry span context information, if a valid span exists.
- Added unit tests for `get_request_and_trace_data`, as well as for both `CloudLoggingHandler` and `StructuredLogHandler`
- Added a system test for Open Telemetry integration using the SDK
- Added `opentelemetry-sdk` as a system test external dependency.


tests/system/test_system.py
```diff
@@ -662,6 +669,40 @@ def test_log_root_handler(self):
...
+            expected_trace_id = f"projects/{Config.CLIENT.project}/traces/{trace.format_trace_id(context.trace_id)}"
+            expected_span_id = trace.format_span_id(context.span_id)
+            expected_tracesampled = context.trace_flags.sampled
+
+            cloud_logger = logging.getLogger(LOGGER_NAME)
+            cloud_logger.addHandler(handler)
+            cloud_logger.warning(LOG_MESSAGE)
+
+            entries = _list_entries(logger)
+            self.assertEqual(len(entries), 1)
```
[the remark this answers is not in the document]
gkevinzheng: Whether or not the instrumentation source entry gets added depends on whether or not the global variable `google.cloud.logging_v2._instrumentation_emitted` is true (see https://github.com/googleapis/python-logging/blob/62641075042a3da9bb9c059d963bad14a1586b1c/google/cloud/logging_v2/__init__.py#L40). When I was running the system tests locally I found that if I ran the entire test suite this case would pass, but if I ran just this test case the instrumentation source entry would be there and it would fail.

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -211,3 +235,37 @@ def get_request_data():
...
+    ) = _parse_current_open_telemetry_span()
+    (
+        http_request,
+        request_trace_id,
+        request_span_id,
+        request_trace_sampled,
+    ) = get_request_data()
+
+    # otel_trace_id existing means the other return values are non-null
+    if otel_trace_id:
```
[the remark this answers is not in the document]
gkevinzheng: If there is no http request data I have it set to null value.

cindy-peng (approved):

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -211,3 +235,37 @@ def get_request_data():
             return http_request, trace_id, span_id, trace_sampled
 
     return None, None, None, False
+
+
+def get_request_and_trace_data():
+    """Helper to get http_request and trace data from supported web
+    frameworks (currently supported: Flask and Django), as well as OpenTelemetry. Attempts
+    to parse trace/spanID from OpenTelemetry first, before going to Traceparent then XCTC.
```
aabmass, line 243: nit this is a bit misleading
```suggestion
    to retrieve trace/spanID from OpenTelemetry first, before going to Traceparent then XCTC.
```

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -191,9 +193,31 @@ def _parse_xcloud_trace(header):
     return trace_id, span_id, trace_sampled
 
 
+def _parse_current_open_telemetry_span():
```
aabmass, line 196: nit i wouldn't use "parse" here

tests/system/test_system.py
```diff
@@ -662,6 +669,40 @@ def test_log_root_handler(self):
...
+            Config.CLIENT, name=handler_name, transport=SyncTransport
+        )
+        # only create the logger to delete, hidden otherwise
+        logger = Config.CLIENT.logger(handler.name)
+        self.to_delete.append(logger)
+
+        # Set up OTel SDK
+        provider = TracerProvider()
+        processor = BatchSpanProcessor(ConsoleSpanExporter())
+        provider.add_span_processor(processor)
```
aabmass, line 687: You can omit these lines if you don't actually want any console output

tests/system/test_system.py
```diff
@@ -662,6 +669,40 @@ def test_log_root_handler(self):
...
+
+        tracer = trace.get_tracer("test_system", tracer_provider=provider)
```
aabmass, line 689: nit this is a bit simpler
```suggestion
        tracer = provider.get_tracer("test_system")
```

tests/unit/handlers/__init__.py
```diff
@@ -11,3 +11,39 @@
...
+    """
+    span_context = opentelemetry.trace.SpanContext(
+        _OTEL_SPAN_CONTEXT_TRACE_ID,
+        _OTEL_SPAN_CONTEXT_SPAN_ID,
+        False,
+        trace_flags=_OTEL_SPAN_CONTEXT_TRACEFLAGS,
+    )
+    span = opentelemetry.trace.NonRecordingSpan(span_context)
+    with mock.patch("opentelemetry.trace.get_current_span", return_value=span) as m:
+        yield m
```
aabmass, line 49: You don't actually need to do any mocking by just setting the span in the real context implementation, see this example https://opentelemetry.io/docs/languages/python/cookbook/#manually-setting-span-context:~:text=%23%20Or%20you%20can,(token)
```suggestion
    ctx = trace.set_span_in_context(span)
    token = context.attach(ctx)
    try:
        yield
    finally:
        context.detach(token)
```

I am guessing that's why you are using `import opentelemetry.trace` instead of `from opentelemetry import trace`

aabmass (approved):
Didn't look too close at tests but LGTM from an OTel perspective

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -211,3 +235,37 @@ def get_request_data():
             return http_request, trace_id, span_id, trace_sampled
 
     return None, None, None, False
+
+
+def get_request_and_trace_data():
```
daniel-sanche, line 240: Does this need to be a new function? The docstrings of `get_request_data` says "Helper to get http_request and trace data from supported web frameworks". It seems to me like this logic should just be merged into the existing one

tests/system/test_system.py
```diff
@@ -662,6 +665,38 @@ def test_log_root_handler(self):
         self.assertEqual(len(entries), 1)
         self.assertEqual(entries[0].payload, expected_payload)
 
+    def test_log_handler_otel_integration(self):
```
daniel-sanche, line 668: Is it possible to also add a system test without the otel sdk imported? 

I'm not exactly sure what that should look like, but I want to make sure we have coverage of the situation where otel isn't used at all

google/cloud/logging_v2/handlers/_helpers.py
gkevinzheng: I was concerned that by changing the function name that it could potentially be a breaking change. It should be OK to change the function name because it's in a private module, right?

tests/system/test_system.py
gkevinzheng: Is it possible to create another file for the newly added system tests, so that the existing TCs aren't importing otel?

google/cloud/logging_v2/handlers/_helpers.py
daniel-sanche: Yes, the module is private, so it should be safe

Does the function need to be renamed though? Why not add new functionality to the existing function?

tests/system/test_system.py
daniel-sanche: Yeah, you can add an extra file. Or just import within the test functions. Or use a test class. Whatever seems cleanest

Does Otel behave any differently based on having the module installed? Or is import state the important part?

gkevinzheng: I actually managed to resolve this by creating a decorator that deletes the otel SDK imports after the test case gets run. I don't think it's perfect but it feels good enough.

daniel-sanche: Ok great! Do you have any test cases that test `_retrieve_current_open_telemetry_span` without otel included?

daniel-sanche (approved):
Consider moving the logic in get_request_and_trace_data into get_request_data. I dont think a new function is required

Other than that, LGTM

Issue #855, opened 2024-02-15
rpc errors only on Python 3.12


efroemling:
Recently I started migrating an App Engine app from Python 3.11 to 3.12 and am seeing occasional cloud logging errors in 3.12 related to cloud logging committing batches in the background thread. The errors don't occur in 3.11 and I don't recall seeing them in prior versions either.

#### Environment details

  - OS type and version: App Engine Standard Environment (which currently seems to be Ubuntu 22.04.3 LTS)
  - Python version: 3.12.1
  - pip version: 23.3.1
  - `google-cloud-logging` version: 3.9.0

#### Steps to reproduce

In my case, switching my app-engine service's runtime from python311 to python312 results in these occasional errors. If I go back to Python 3.11 the errors cease.

#### Code example

The following is the entirety of my logging setup. I'm mostly relying on standard Python log calls but have one or two places where I talk to cloud logging directly in case that could be a culprit.
```python
        # Wire up to handle standard Python logs.
        gc_log_client = google.cloud.logging.Client()
        gc_log_client.setup_logging(log_level=logging.INFO, name='bamaster')

        # Also set up a direct logger (used in a few places only).
        gc_logger_direct = gc_log_client.logger(name='bamaster_direct')
```

#### Stack trace
The following are the errors I'm seeing. The stack traces seem to be oddly cut off.
<img width="1206" alt="Screenshot 2024-02-15 at 10 31 58 AM" src="https://github.com/googleapis/python-logging/assets/10857610/38b9274c-ac41-4d6e-91cf-41c3697d357d">

Any ideas what could have changed here between Python 3.11 and 3.12?
Please let me know if there is anything I can do to provide more info on this. Thanks.

gkevinzheng:
@efroemling I can't see the rest of the stack trace there, but by any chance does it look like this:

```
File "/usr/local/google/home/kevzheng/.pyenv/versions/3.12.0/lib/python3.12/site-packages/google/api_core/gapic_v1/method.py", line 131, in __call__
    return wrapped_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/kevzheng/.pyenv/versions/3.12.0/lib/python3.12/site-packages/google/api_core/retry.py", line 372, in retry_wrapped_func
    return retry_target(
           ^^^^^^^^^^^^^
[...]

PR #917, opened 2024-07-10
feat: Added flushes/close functionality to logging handlers

gkevinzheng:
For design doc see [go/python-logging-background-thread](http://goto.google.com/python-logging-background-thread)


google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -213,6 +215,11 @@ def emit(self, record):
         labels = {**add_resource_labels(resource, record), **(labels or {})} or None
 
         # send off request
+        if self.transport is None:
+            self.transport = self._transport_cls(
+                self.client, self.name, resource=self.resource
+            )
```
daniel-sanche, line 221: Wouldn't this mean the handler has been closed? Why not just raise an error, instead of re-creating the transport?

google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -225,6 +232,20 @@ def emit(self, record):
...
+        For SyncTransport, this is a no-op.
+        """
+        super(CloudLoggingHandler, self).flush()
+        if self.transport is not None:
+            self.transport.flush()
+
+    def close(self):
+        """Closes the log handler and cleans up all Transport objects used."""
+        self.transport.close()
+        self.transport = None
```
daniel-sanche, line 247: I don't like the idea of using `transport == None` as a way to check if the handler is closed. That makes the types quite a bit more complicated.

Can't we use a new flag for this?

google/cloud/logging_v2/handlers/transports/background_thread.py
```diff
@@ -38,6 +38,12 @@
 _WORKER_TERMINATOR = object()
 _LOGGER = logging.getLogger(__name__)
 
+_CLOSE_THREAD_SHUTDOWN_ERROR_MSG = (
+    "CloudLoggingHandler shutting down, cannot send logs entries to Cloud Logging due to "
+    "inconsistent threading behavior at shutdown. To avoid this issue, flush the logging handler "
+    "manually or switch to StructuredLogHandler."
```
daniel-sanche, line 44: Maybe mention client.close or handler.close too?

google/cloud/logging_v2/handlers/transports/background_thread.py
```diff
@@ -202,26 +210,40 @@ def stop(self, *, grace_period=None):
...
 
-    def _main_thread_terminated(self):
+    def _close(self):
         """Callback that attempts to send pending logs before termination."""
         if not self.is_alive:
             return
 
+        # Print different messages to the user depending on whether or not the
+        # program is shutting down. This is because this function now handles both
+        # the atexit handler and the regular close.
```
daniel-sanche, line 220: Can we decouple this? Maybe let the caller pass in a message or something?

google/cloud/logging_v2/handlers/handlers.py
gkevinzheng: I based this off of some logging handlers I saw in the Python standard library, which will reopen a closed connection if that connection is closed.

daniel-sanche: Personally I think this adds extra complexity, so I'd prefer to keep them permanently closed if we can get away with that

But if you think this behavior would help us integrate with the standard library, it makes sense to keep it.

otorreno:
I am not fully sure this will solve #850 and #855. I see 2 possible scenarios:
1. A user calls flush on the logging client and then the program exit. That would call the _close function which would either attempt to send what is in the queue if the main thread has not terminated or it will just not send whatever is in the queue.
2. Even if a user would call flush and then close, nothing guarantees that between these 2 calls some message got to the queue. I think there will be similar outcomes to the previous point.

Have you considered flushing the queue at termination? Something like the following (conceptually):
1. Mark as shutting down (nothing else gets queued
2. Flush the queue
3. Mark as terminated

I wonder if the atexit hook should remain. I believe the main thread will be in a state where is_alive may return true, but underneath it would still error out because it can’t create new threads (this is a guess). What do you think about letting users explicitly calling the teardown/close?

gkevinzheng:
> I am not fully sure this will solve #850 and #855. I see 2 possible scenarios:
> 
> 1. A user calls flush on the logging client and then the program exit. That would call the _close function which would either attempt to send what is in the queue if the main thread has not terminated or it will just not send whatever is in the queue.
> 2. Even if a user would call flush and then close, nothing guarantees that between these 2 calls some message got to the queue. I think there will be similar outcomes to the previous point.
> 
> Have you considered flushing the queue at termination? Something like the following (conceptually):
> 
> 1. Mark as shutting down (nothing else gets queued
> 2. Flush the queue
> 3. Mark as terminated
> 
> I wonder if the atexit hook should remain. I believe the main thread will be in a state where is_alive may return true, but underneath it would still error out because it can’t create new threads (this is a guess). What do you think about letting users explicitly calling the teardown/close?

Part of this PR is also adding functionality to explicitly let users call teardown/close for clients. Flushing the queue at termination is not an option because in order to send log entries over the wire to Cloud Logging, gRPC threads are needed, which we cannot create new ones at shutdown, causing the two bugs.

google/cloud/logging_v2/handlers/handlers.py
```diff
@@ -188,7 +188,10 @@ def __init__(
             resource = detect_resource(client.project)
         self.name = name
         self.client = client
+        client._handlers.add(self)
         self.transport = transport(client, name, resource=resource)
+        self.transport_open = True
```
daniel-sanche, line 193: Can this be made private?

daniel-sanche (changes requested):

daniel-sanche (approved):

PR #925, opened 2024-08-01
test: Added timestamp to sink names + autodelete sinks older than 2 hours in export_test.py

### LINKED ISSUE #923: Clean up samples resources
Samples presubmits are failing at HEAD with ` google.api_core.exceptions.ResourceExhausted: 429 Quota limit of logging.googleapis.com/log_sinks_count reached`. As an example, see this [build log](https://btx.cloud.google.com/invocations/f17f72dc-ae5d-4a86-a8b4-7c7a8edc027e/log).

It's possible that resources created during the samples test may not be fully cleaned up. For example, see the code below where a resource may not be deleted if an exception is raised when calling `delete`

https://github.com/googleapis/python-logging/blob/5bc9b448506c0c51d20ea2b475b5366afe921d05/samples/snippets/export_test.py#L72-L77

gkevinzheng:
Added a timestamp field in `TEST_SINK_NAME_IMPL` and a cleanup threshold so that the `export_test.py` test suite deletes sinks older than 2 hours when run.

Also added exponential backoff to sink deletion to reduce the chances of sink deletion failure.

Fixes #923

### DIFF
--- a/samples/snippets/export_test.py
+++ b/samples/snippets/export_test.py
@@ -35,16 +35,17 @@ def _random_id():
     )
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope="module")
 def cleanup_old_sinks():
     client = logging.Client()
-    test_sink_name_regex = TEST_SINK_NAME_TMPL.format('[A-Z0-9]{6}$')
+    test_sink_name_regex = TEST_SINK_NAME_TMPL.format("[A-Z0-9]{6}$")
     for sink in client.list_sinks():
         if re.match(test_sink_name_regex, sink.name):
             try:
                 sink.delete()
             except Exception:
                 pass
+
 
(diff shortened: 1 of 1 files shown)


samples/snippets/export_test.py
```diff
@@ -48,10 +79,7 @@ def example_sink():
 
     yield sink
 
-    try:
-        sink.delete()
-    except Exception:
-        pass
+    _delete_sink(sink)
```
parthea, line 82: We should handle the exception otherwise we will have flaky tests. Applies throughout

parthea (approved):

samples/snippets/export_test.py
parthea: My bad. It looks like this is already being done

PR #932, opened 2024-08-08
chore: Added mypy to noxfile

gkevinzheng:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass


owlbot.py
```diff
@@ -265,3 +265,22 @@ def place_before(path, text, *before_text, escape=None):
                     file_path,
                     merge=lambda src, dst, _,: f"{dst}\n{src}",
                 )
+
+place_before(
+    "noxfile.py",
+    """
```
daniel-sanche, line 271: I see [firestore](https://github.com/googleapis/python-firestore/blob/a0d2d7e26bd2008cc1f19727002a8c663bce78ec/owlbot.py#L299) adds an escape character here. Maybe that's why it's not working?

gkevinzheng: I found out that using `s.replace` instead of `place_before` makes it work. The escape characters were used because `s.replace` takes in a regex.

noxfile.py
```diff
@@ -134,6 +134,15 @@ def format(session):
     )
 
 
+@nox.session(python=DEFAULT_PYTHON_VERSION)
+def mypy(session):
+    """Verify type hints are mypy compatible."""
+    session.install("-e", ".")
+    session.install("mypy", "types-setuptools")
+    # TODO: also verify types on tests and logging_v2, big undertakings
```
daniel-sanche, line 142: are those big undertakings? logging_v2 should already have valid types in place. And I'd say tests are optional

daniel-sanche:
you should also add a github action to execute mypy as a presubmit check

PR #946, opened 2024-10-11
fix: 16-bit hexadecimal formatting for XCTC span IDs

### LINKED ISSUE #942: Span ID reported when "X-Cloud-Trace-Context" is used is incorrect
For Googlers: motivating details / context can be found in Buganizer issue # 367762417 

#### Steps to reproduce

  1. Execute in an environment in which tracing context has been supplied via `X-Cloud-Trace-Context`.

#### Observed

  The `LogEntry` that is created contains a `span_id` that is incorrect.
  
  Note that `LogEntry.span_id` should be the hexadecimal encoding of the span ID, whereas `X-Cloud-Trace-Context` (unlike `traceparent`) encodes the `span_id` in decimal format.

#### Expected

 The `LogEntry.span_id` field should be populated with the canonical (hexadecimal) representation of the span ID.

gkevinzheng:
Fixes #942 

Also updated link to comment for XCTC formatting, as that seems to have been moved prior.

### DIFF
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -199,15 +199,14 @@ def _parse_xcloud_trace(header):
             trace_sampled = match.group(5) == "1"
 
             # Convert the span ID to 16-bit hexadecimal instead of decimal
-            if span_id is not None:
-                try:
-                    span_id_int = int(span_id)
-                    if span_id_int > 0 and span_id_int < 2**64:
-                        span_id = f"{span_id_int:016x}"
-                    else:
-                        span_id = None
-                except ValueError:
+            try:
+                span_id_int = int(span_id)
+                if span_id_int > 0 and span_id_int < 2 ** 64:
+                    span_id = f"{span_id_int:016x}"
+                else:
                     span_id = None
+            except (ValueError, TypeError):
+                span_id = None
 
         except IndexError:
             pass
--- a/google/cloud/logging_v2/handlers/_helpers.py
+++ b/google/cloud/logging_v2/handlers/_helpers.py
@@ -174,13 +174,22 @@ def _parse_xcloud_trace(header):
     Args:
         header (str): the string extracted from the X_CLOUD_TRACE header
     Returns:
-        Tuple[Optional[dict], Optional[str], bool]:
+        Tuple[Optional[str], Optional[str], bool]:
             The trace_id, span_id and trace_sampled extracted from the header
             Each field will be None if not found.
     """
     trace_id = span_id = None
     trace_sampled = False
-    # see https://cloud.google.com/trace/docs/setup for X-Cloud-Trace_Context format
+
+    # As per the format described at https://cloud.google.com/trace/docs/trace-context#legacy-http-header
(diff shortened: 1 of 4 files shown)


google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -174,20 +174,41 @@ def _parse_xcloud_trace(header):
...
             trace_sampled = match.group(5) == "1"
+
+            # Convert the span ID to 16-bit hexadecimal instead of decimal
+            if span_id is not None:
+                try:
+                    span_id_int = int(span_id)
+                    if span_id_int > 0 and span_id_int < 2**64:
+                        span_id = f"{span_id_int:016x}"
+                    else:
+                        span_id = None
```
daniel-sanche, line 208: nit: maybe raise an exception here, since we already have the try/catch?

```
try:
    span_id_int = int(span_id)
    if span_id_int < 0 or span_id_int > 2**64:
        raise ValueError("span_id outside valid range")
    span_id = f"{span_id_int:016x}"
except ValueError:
    span_id = None
```

daniel-sanche (approved):

google/cloud/logging_v2/handlers/_helpers.py
```diff
@@ -174,20 +174,41 @@ def _parse_xcloud_trace(header):
...
         try:
             regex = r"([\w-]+)?(\/?([\w-]+))?(;?o=(\d))?"
             match = re.match(regex, header)
             trace_id = match.group(1)
             span_id = match.group(3)
             trace_sampled = match.group(5) == "1"
+
+            # Convert the span ID to 16-bit hexadecimal instead of decimal
+            if span_id is not None:
+                try:
```
michaelsafyan: You have a lot of layers of nesting here... consider splitting this up into multiple smaller helper functions. For example:

```
   xctc_trace_id, xctc_span_id, xctc_options = _parse_xct_header(header)
   trace_id = _validate_and_normalize_xctc_trace_id(xctc_trace_id)
   span_id = _validate_and_normalize_xctc_span_id(xctc_span_id)
   sampled = _extract_should_sample_from_xctc_option(xctc_options)
```

google/cloud/logging_v2/handlers/_helpers.py
michaelsafyan: If parts of the header fail to validate, you may want to consider making it more obvious that this has happened. Consider some form of debug logging or other tracking if not propagated an error to the caller.

tests/unit/handlers/test__helpers.py
```diff
@@ -501,43 +507,62 @@ def test_no_span(self):
...
         self.assertEqual(span_id, expected_span)
         self.assertEqual(sampled, False)
 
+    def test_with_span_decimal_not_in_bounds(self):
+        input_spans = ["0", "9" * 100]
+
+        for input_span in input_spans:
+            expected_trace = "12345"
+            input_span = "67890"
+            expected_span = "10932".zfill(16)
```
michaelsafyan, line 534: Why is this the expected span when the input span is invalid? Shouldn't the expected span be None?

Issue #943, opened 2024-09-23
Crash with google.protobuf.json_format.ParseError: Value '2024-09-23' has unexpected type <class 'datetime.date'>


daaain:
#### Environment details

  - OS type and version: `python:3.11.9-bookworm` base image
  - Python version: `3.11.9`
  - pip version: `pip 24.0 from /usr/local/lib/python3.11/site-packages/pip (python 3.11)`
  - `google-cloud-logging` version: `3.11.2`

#### Steps to reproduce

Trying to send down a dict that has a DateTime property blows the library up with a `google.protobuf.json_format.ParseError`

I guess I could work around it manually and convert it to a string myself, but the bigger issue is that I wouldn't expect a logging library to crash my application.

Ideally the library should handle anything thrown at it with a best effort string conversion (even if it's just a useless class instance identity or anything), but at the very least have a way to log its own error rather than throwing an unhandled exception? Or are end users expected to wrap it in try / except and handle certain errors? Or is there a way to give the library a custom JSON parser?

#### Code example

```python
from datetime import datetime
from google.cloud import logging

logging_client = logging.Client()
logging_client.setup_logging()

my_dict = {"my_date": datetime.now()}

logging_client.logger("default").log_struct({"message": "message", "my_dict": my_dict}, severity="INFO")
```

I thought maybe using the `logging` std lib might make a difference but that also crashes:

```python
import logging
from datetime import datetime
from google.cloud import logging as gcloud_logging

logging_client = gcloud_logging.Client()
logging_client.setup_logging()

data_dict = {"my_date": datetime.now()}
logging.info("message field", extra={"json_fields": data_dict})

>>> Failed to submit 1 logs.
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/google/protobuf/json_format.py", line 600, in _ConvertFieldValuePair
    self.ConvertMessage(value, sub_message, '{0}.{1}'.format(path, name))
  File "/usr/local/lib/python3.11/site-packages/google/protobuf/json_format.py", line 485, in ConvertMessage
    methodcaller(_WKTJSONMETHODS[full_name][1], value, message, path)(self)
  File "/usr/local/lib/python3.11/site-packages/google/protobuf/json_format.py", line 704, in _ConvertStructMessage
    self._ConvertValueMessage(value[key], message.fields[key],
  File "/usr/local/lib/python3.11/site-packages/google/protobuf/json_format.py", line 682, in _ConvertValueMessage
[...]

PR #958, opened 2025-01-06
fix: Made `write_entries` raise `ValueError` on `ParseError`s

gkevinzheng:
Two fixes for #943:

1. Made `write_entries` raise `ValueError` on invalid log entries.
2. Added documentation clarifying the expected format of `log_struct` and clarifying the `ValueError` that gets raised.

### DIFF
--- a/google/cloud/logging_v2/logger.py
+++ b/google/cloud/logging_v2/logger.py
@@ -206,12 +206,22 @@ class Logger(object):
         See
         https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write
 
-        The dictionary entry must be able to be serializable to a Protobuf Struct
-        (see https://protobuf.dev/reference/protobuf/google.protobuf/#value for more
-        details), otherwise it will not be logged, and a :class:`ValueError` will be raised.
+        The message must be able to be serializable to a Protobuf Struct.
+        It must be a dictionary of strings to one of the following:
+
+            - :class:`str`
+            - :class:`int`
+            - :class:`float`
+            - :class:`bool`
+            - :class:`list[str|float|int|bool|list|dict|None]`
+            - :class:`dict[str, str|float|int|bool|list|dict|None]`
+
+        For more details on Protobuf structs, see https://protobuf.dev/reference/protobuf/google.protobuf/#value.
+        If the provided dictionary cannot be serialized into a Protobuf struct,
+        it will not be logged, and a :class:`ValueError` will be raised.
 
         Args:
-            info (dict[str|float|int|bool|list|dict|None]):
+            info (dict[str, str|float|int|bool|list|dict|None]):
                 the log entry information.
             client (Optional[~logging_v2.client.Client]):
                 The client to use.  If not passed, falls back to the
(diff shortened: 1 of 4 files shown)


google/cloud/logging_v2/logger.py
```diff
@@ -162,11 +164,15 @@ def _do_log(self, client, _entry_class, payload=None, **kw):
...
+
         if google.cloud.logging_v2._instrumentation_emitted is False:
             entries = _add_instrumentation(entries, **kw)
             google.cloud.logging_v2._instrumentation_emitted = True
         # partial_success is true to avoid dropping instrumentation logs
-        client.logging_api.write_entries(entries, partial_success=True)
+        try:
+            client.logging_api.write_entries(entries, partial_success=True)
+        except Exception:
+            _LOGGER.error("Failed to submit log message.", exc_info=True)
```
daniel-sanche, line 175: I don't think it's a good idea to just swallow all exceptions here. All grpc networking errors would now be ignored (ex bad IAM credentials), and users will have no idea anything is wrong unless they know to subscribe to the internal _LOGGER. And existing customers may have error handling logic in place that will now be bypassed, so this could be a breaking change hidden in a bug fix

google/cloud/logging_v2/logger.py
```diff
@@ -206,7 +212,11 @@ def log_struct(self, info, *, client=None, **kw):
         https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write
 
         Args:
-            info (dict): the log entry information
+            info (dict):
+                the log entry information. The dict must be serializable
+                to a Protobuf Struct (map from strings to Values, see
+                https://protobuf.dev/reference/protobuf/google.protobuf/#value),
```
daniel-sanche, line 218: is it possible to lay out the specific types that are serializable? (maybe not if we support nesting dicts, but just wanted to check)

google/cloud/logging_v2/logger.py
```diff
@@ -206,7 +212,11 @@ def log_struct(self, info, *, client=None, **kw):
...
+                otherwise the item will not be logged.
```
daniel-sanche, line 219: it might be good to add more info to the main text too, instead of "Log a dictionary message"

google/cloud/logging_v2/logger.py
```diff
@@ -468,11 +478,12 @@ def commit(self, *, client=None, partial_success=True):
             client.logging_api.write_entries(
                 entries, partial_success=partial_success, **kwargs
             )
-        except InvalidArgument as e:
-            # InvalidArgument is often sent when a log is too large
-            # attempt to attach extra contex on which log caused error
-            self._append_context_to_error(e)
-            raise e
+        except Exception as e:
```
daniel-sanche, line 481: same comment as above: hiding exceptions by default seems like it could have unintentional consequences, and could be a breaking change

daniel-sanche (changes requested):

google/cloud/logging_v2/logger.py
daniel-sanche: You swallowing catch the specific exception instead of catching all exceptions?

Honestly, I think just raising an exception here makes sense, so the user knows their input isn't handled. I'd do a `raise ValueError(...) from e` though, so we can give more context instead of raising the raw internal error. And then add a `Raises:` section in the docstring describing this situation

google/cloud/logging_v2/logger.py
```diff
@@ -200,18 +201,28 @@ def log_text(self, text, *, client=None, **kw):
...
     def log_struct(self, info, *, client=None, **kw):
-        """Log a dictionary message
+        """Logs a dictionary message.
 
         See
         https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write
 
+        The dictionary entry must be able to be serializable to a Protobuf Struct
+        (see https://protobuf.dev/reference/protobuf/google.protobuf/#value for more
+        details), otherwise it will not be logged.
```
daniel-sanche, line 211: "Otherwise a ValueError exception will be raised?"

Also, maybe include that this essentially means strings, floats, bools, lists, and dicts, and None?

daniel-sanche (approved):

